PR #2, opened 2018-01-08
Adds ;diskus script to cast disks on people if they don't have one

sandersch:
A simple script that I use to cast disks for people.


ondreian:
thanks!

PR #3, opened 2018-03-20
Add ;absorb utility for experience pulses

sandersch:
Experience pulse utility script

By default, this script will run in the background and announce experience gains as they are absorbed.

Also supports some actions around pulses and mind saturation:

  * Run until experience absorbed and then exit (ala ;pulse.lic)
  * Run a command or start another script on an experience pulse (ala ;errand.lic)
  * Log off when mind clears or reaches a specified level

### DIFF
--- a/scripts/absorb.lic
+++ b/scripts/absorb.lic
@@ -1,0 +1,128 @@
+=begin
+  version: 1.0
+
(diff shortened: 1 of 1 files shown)


ondreian:
Thanks

PR #5, opened 2018-04-18
;notify tweaks

sandersch:
Some minor changes to ;notify that I have been meaning to submit for a bit for discussion. These three commits should all be independent from each other so we can take out/change anything you don't like.

The Regexp.union change has almost the exact same behavior as your existing monkey patch, but I try to minimize the use of those if there is a built in way of doing it. The Regexp method also preserves flags like `/ignore case/i` when combining regexs.

I updated the help to include the play command but also made the script try to play a sound when changing it as this is the behavior I expected.


ondreian:
Thanks

PR #6, opened 2018-04-18
use a tcp message sink;

ondreian:
1. Updates notify to create a background TCP server that accepts messages from all clients and forwards them to the system notification library
2. Uses a hybrid TTL/LRU cache mechanism to rate limit messages from clients

### DIFF
--- a/scripts/notify.lic
+++ b/scripts/notify.lic
@@ -30,7 +30,7 @@
     # because it is in a different thread
     Notify::Sounds.set "zelda/email"
   
-    Notify.notify(from: "taters", body: "i'm important", type: Script.current.name)
+    Notify.notify("Ondreian", from: "taters", body: "i'm important", type: Script.current.name)
 
   notes:
     * to be able to use Notify from another script it must be ;trusted
@@ -49,6 +49,7 @@
 =end
 require "ostruct"
 require "socket"
+require "json"
 
 class String
   def is_i?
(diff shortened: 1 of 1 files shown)


scripts/notify.lic
```diff
@@ -160,12 +440,12 @@ module Notify
   end
```
sandersch: `title = [message.from, Char.name].join(" => ")` is showing all notification as being "to" the character running the server thread. We probably need to save this onto `message` in `fmt_message` and update the title here to use `message.to`

ondreian: Good catch.  I pushed a commit that passes the real `Char.name` over tcp using `json` now

scripts/notify.lic
```diff
@@ -30,7 +30,7 @@
     # because it is in a different thread
     Notify::Sounds.set "zelda/email"
   
-    Notify.notify(from: "taters", body: "i'm important", type: Script.current.name)
+    Notify.notify("Ondreian", from: "taters", body: "i'm important", type: Script.current.name)
```
sandersch: I'm not really liking this API change so far. First of all, it is a breaking so all of my existing uses are borked. This is unfortunate, but not a deal breaker. Secondly, I have yet to want to notify from a character other than the one making the call. Is there any way we could convert this into an optional field and fallback to `Char.name` to preserve the existing API?

ondreian: The problem is, it's used by the tcp sink also, but i agree, a key argument makes more sense here that defaults to the current character makes more sense, i will fix

PR #7, opened 2018-04-22
Added new flee option

mrhoribu:
Added option to flee from webs as well as added 917 to not stance offensive when using incant option

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -5,7 +5,7 @@
         hunting script for hunting
  
               author: elanthia-online
-        contributers: SpiffyJr, Tillmen, Kalros, Hazado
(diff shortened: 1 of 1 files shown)


sandersch:
Looks good to me :shipit:

scripts/bigshot.lic
```diff
@@ -4,35 +4,38 @@
  
         hunting script for hunting
  
-              author: nobody
+              author: elanthia-online
         contributers: SpiffyJr, Tillmen, Kalros, Hazado
```
sandersch: You should add yourself as a contributor especially since you are the de facto maintainer at this point.

mrhoribu: good point.

PR #9, opened 2018-04-25
Initial version of ;rofl-puzzles.lic

sandersch:
### Rings of Lumnis Puzzle Solver

Usage:
  `;rofl-puzzles <puzzle name>`

where `<puzzle name>` is one of the following:

    auto       automatically solve puzzles when they appear (default behavior)

    statue     move some some statues around on an altar
    mosaic     find the right gem and symbol to put into the mosaic
    crystal    choose the correct damage type to destroy monster paintings
    stars      select the correct constellation to place in the night sky
    ghost      use a wand to do stuff to a ghost
    scramble   guess the word from scrambled letters
    symbol     paint an Arkati symbol the correct colors and assemble
    box        get some stuff from a box and do some things
    candle     light the correct candle

  contributions/thanks:

  * Alastir for creating `;RoL` to answer trivia questions and providing solutions to many anagrams and the ghost, box, and stars puzzles
  * Cigger for the gem anagram answers and testing
  * Naamit and Maetriks for creature anagrams
  * Claudaro for help testing and providing feedback
  * Roblar for some hints to puzzle solutions
  * Many other people in the HOA #puzzle_spoilers channel for answers and feedback

### DIFF
--- a/scripts/rofl-puzzles.lic
+++ b/scripts/rofl-puzzles.lic
@@ -1,0 +1,851 @@
+=begin
+  version: 0.1
+
+  Rings of Lumnis Puzzle Solver
+
+  Usage:
+    ;rofl-puzzles <puzzle name>
+
+  where <puzzle name> is one of the following:
+
+    auto       automatically solve puzzles when they appear (default behavior)
+
+    statue     move some some statues around on an altar
+    mosaic     find the right gem and symbol to put into the mosaic
+    crystal    choose the correct damage type to destroy monster paintings
+    stars      select the correct constellation to place in the night sky
+    ghost      use a wand to do stuff to a ghost
+    scramble   guess the word from scrambled letters
+    symbol     paint an Arkati symbol the correct colors and assemble
+    box        get some stuff from a box and do some things
+    candle     light the correct candle
+
+  contributions/thanks:
+    * Alastir for creating ;RoL to answer trivia questions and providing
+      solutions to many anagrams and the ghost, box, and stars puzzles
+    * Cigger for the gem anagram answers and testing
+    * Naamit and Maetriks for creature anagrams
+    * Claudaro for help testing and providing feedback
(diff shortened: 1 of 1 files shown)


scripts/rofl-puzzles.lic
```diff
@@ -0,0 +1,851 @@
...
+
+  **Pull Requests Welcome** - you can contribute changes at:
+  https://github.com/elanthia-online/scripts/tree/master/scripts/rofl-puzzles.lic
+
+  Author: Kragdruk
+  Tags: festival, RoL, Rings
+=end
+
+module RofL
+  class Statue
```
sandersch: I would have really preferred a way to split this up into separate files for each puzzle class. For some of my personal scripts that I haven't published, I've worked around this using load `./other_file.lic` but that requires `;trust`ing all the things which is not worth it for stuff I want to share in the repo.

scripts/rofl-puzzles.lic
```diff
@@ -0,0 +1,851 @@
...
+  respond
+  respond "    %-10s %s" % ["auto", "automatically solve puzzles when they appear (default behavior)"]
+  respond
+  RofL::PUZZLES.each do |puzzle|
+    respond "    %-10s %s" % [puzzle.name, puzzle.description]
+  end
+  respond
+end
+
+if defined?(Lich)
```
sandersch: This is my stupid ruby trick to facilitate testing. I can `load "./rofl-puzzles.lic` from the test file and it won't execute any of the code in here. I will probably make a separate issue about testing in general because I like to write specs for stuff when possible, but I also realize that almost no one else cares to do so.

ondreian: I love testing, and have plans for creating a mock CI environment we can use, just not time to implement it right now.  Maybe i can just sketch my ideas out in an issue

PR #12, opened 2018-04-30
added cold as an element type

mrhoribu:
updated the regex for cmd_spell to have cold in it for the extra variable tag. incremented bigshot version accordingly.


sandersch:
I would squash these into a single commit, but not a big deal to me if you don't.

PR #13, opened 2018-05-16
;wps: Smithy helper script v1.0

sandersch:
This script will request multiple services from the smithy. It tries to be strict about what it will accept for input and to verify that a blacksmith NPC is present before requesting services.

This is kind of overkill for a script that provides minimal utility over `;e count.times { ... }`, but I wanted to ride out the command pattern and some simple extract method refactoring to see where it landed. Welcome any feedback if you have any as I have a couple points for discussion.

### DIFF
--- a/scripts/wps.lic
+++ b/scripts/wps.lic
@@ -1,0 +1,138 @@
+=begin
+
+  WPS smithy service helper script
+
+  This script will request multiple services from the smithy. It tries to be
+  strict about what it will accept for input and to verify that a blacksmith
+  NPC is present before requesting services.
+
+  Usage:
+
+  ;wps <count> <crit, damage, or sighting>
+
+  author: Kragdruk
+  version: 1.0
+  tags: util, ccf, duskruin, festival, merchant, wps, smithy
+
+=end
+
+module WPS
+  class Error < StandardError
+    def for_cmdline
+      <<-OUTPUT
+
+    ERROR: #{message}
+
+      OUTPUT
+    end
+  end
+
+  class ArgumentError < Error; end
+
+  SERVICE_TYPES = %w[critical damage sighting]
+
+  def self.determine_service_type_from(arg)
+    SERVICE_TYPES.find { |t| t.start_with? arg.downcase }
+  end
+
+  BLACKSMITH_DESCRIPTIONS = [
+    "old dwarven blacksmith",
+  ]
+
+  def self.blacksmith
+    GameObj.npcs.find do |npc|
+      BLACKSMITH_DESCRIPTIONS.include? npc.name
+    end
+  end
+
+  def self.print_usage
+    respond <<-USAGE
+    Usage:
+
+    #{$clean_lich_char}#{Script.current.name} <count> <crit, damage, or sighting>
+
+    USAGE
+  end
+
+  def self.perform_services(opts)
+    opts.fetch(:count).times do
+      perform_service_request(opts.fetch(:service))
+      perform_service_confirmation
+    end
+  end
+
+  def self.perform_service_request(service_type)
+    if result = dothistimeout("ask ##{blacksmith.id} about #{service_type}", 5, /Modified \w+ Type: #{service_type}/i)
+      result
+    else
+      fail Error.new("doesn't seem that we were able to ask the blacksmith for work")
+    end
+  end
+
+  def self.perform_service_confirmation
+    if result = dothistimeout( "ask ##{blacksmith.id} about confirm", 5, /quickly returns, idly polishing it with a dirty rag as it hands it back to you/)
+      result
+    else
+      fail Error.new("doesn't seem that we were able to ask the blacksmith for work")
+    end
+  end
+
+  def self.validate_script_arguments
+    Script.current.vars[2] or
+      fail WPS::ArgumentError.new("Not enough script arguments given")
+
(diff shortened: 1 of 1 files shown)


scripts/smithy.lic
```diff
@@ -0,0 +1,137 @@
...
+=end
+
+module Smithy
+  SERVICE_TYPES = %w[critical damage sighting]
+
+  def self.determine_service_type_from(arg)
+    SERVICE_TYPES.find { |t| t.start_with? arg.downcase }
+  end
+
+  BLACKSMITH_DESCRIPTIONS = [
```
sandersch: I set it up like this in case the DR and CCF blacksmiths had different descriptions, but I checked my logs and they are the exact same. The cost of abstraction is minimal here and I also abused it for some simple testing, so figured I'd keep it like this even though there is only one entry in the list.

scripts/smithy.lic
```diff
@@ -0,0 +1,137 @@
...
+  ;smithy <count> <crit, damage, or sighting>
+
+  author: Kragdruk
+  version: 1.0
+  tags: util, ccf, duskruin, festival, merchant
+
+=end
+
+module Smithy
+  SERVICE_TYPES = %w[critical damage sighting]
```
sandersch: I haven't ever actually paid for damage weighting/padding or sighting, but I verified that the requesting works the same.

scripts/smithy.lic
```diff
@@ -0,0 +1,137 @@
...
+  strict about what it will accept for input and to verify that a blacksmith
+  NPC is present before requesting services.
+
+  Usage:
+
+  ;smithy <count> <crit, damage, or sighting>
+
+  author: Kragdruk
+  version: 1.0
+  tags: util, ccf, duskruin, festival, merchant
```
sandersch: Do we have any thoughts on tags or some kind of taxonomy? I have been looking at existing tags in `;repo gui` to get some ideas but would be open to more consistency for these.

scripts/smithy.lic
```diff
@@ -0,0 +1,137 @@
...
+
+  def self.normalize_input(count_input, service_input)
+    {}.tap do |opts|
+      opts[:count] = count_input.to_i
+      opts[:service] = determine_service_type_from service_input
+    end
+  end
+
+  def self.run
+    if validate
```
sandersch: `validate` should bail on any failures and never return a falsey value here, but figured it could prevent us from doing the wrong thing if I forgot to add an exit to one of the validation failure handlers. I also think it is more intention revealing especially to newcomers than e.g. naming the method `validate!`.

scripts/smithy.lic
```diff
@@ -0,0 +1,139 @@
...
+      respond
+      pause_script
+
+      perform_services(opts)
+    end
+  end
+end
+
+if defined? Lich
+  Smithy.run
```
sandersch: I originally snuck this in at the end of the class definition without the Lich guard like
```
module Smithy
  ...

  run
end
```

but I thought that was kind of unintuitive.

scripts/smithy.lic
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  def self.perform_service_request(service_type)
+    if result = dothistimeout("ask ##{blacksmith.id} about #{service_type}", 5, /Modified \w+ Type: #{service_type}/i)
+      result
+    else
+      respond
+      respond "  ERROR: doesn't seem that we were able to ask the blacksmith for work"
+      respond
+      exit
```
sandersch: I considered making a `fatal` helper method for this repeated idiom and/or switching to exceptions, but didn't feel strongly enough either.

scripts/smithy.lic
```diff
@@ -0,0 +1,139 @@
...
+      BLACKSMITH_DESCRIPTIONS.include? npc.name
+    end
+  end
+
+  def self.print_usage
+    respond
+    respond "  Usage:"
+    respond
+    respond "  #{$clean_lich_char}#{Script.current.name} <count> <crit, damage, or sighting>"
+    respond
```
sandersch: How do you guys like to format your script's output? I use a lot of `echo` in my personal scripts because it is helpful for attribution when debugging, but it is kind of ugly for normal output during regular operation. How do you handle warnings/errors? My only real best practice here is to avoid `puts` because stdout is connected to the main lich process instead of the frontend when lich is running in detachable-client mode.

mrhoribu: I prefer `echo` over `respond` for same reason that it shows the script that sent the info.

mrhoribu:
Only issue is there's another ;smithy on the repo by Drafix right now, so would either we rename this or we get permission from Drafix to incorporate whatever functionality that script has as well, then Drafix would remove his version and this could replace it.

scripts/smithy.lic
ondreian: Probably should have `wps` as a tag at the least

scripts/smithy.lic
ondreian: I prefer exceptions, because you get free stacktraces to help knowledgeable people debug

scripts/smithy.lic
ondreian: Heredocs make for much better Multiline output

sandersch:
I thought I checked for a name collision but apparently not. Will update with a different name, probably `;wps`.

PR #15, opened 2018-05-18
[WIP] Cluster coordination util

ondreian:
This is a rough draft for a redis-backed Cluster coordination utility service.

### DIFF
--- a/scripts/cluster.lic
+++ b/scripts/cluster.lic
@@ -100,7 +100,7 @@ module Cluster
   ## invoke a given method on a Node within an available Actor
   ##
   def self.call(id, method, *args, actor: :bridge)
-    if node = Cluster.get(id)
+    if node = Cluster.node(id)
       node[actor].send(method, *args)
     end
   end
--- a/scripts/cluster.lic
+++ b/scripts/cluster.lic
@@ -1,0 +1,179 @@
+require "dcell"
+
+module Cluster
+  ##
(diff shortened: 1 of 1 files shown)


scripts/cluster.lic
```diff
@@ -0,0 +1,179 @@
...
+    DCell::Node.all.map do |node|
+      [node.id, 
+        !node[actor].nil?]
+    end
+  end
+  ##
+  ## invoke a given method on a Node within an available Actor
+  ##
+  def self.call(id, method, *args, actor: :bridge)
+    if node = Cluster.get(id)
```
sandersch: I was getting: `Exception: private method `get' called for Cluster:Module` when using this, but it seems like it is working after I updated it to: `if node = Cluster.node(id)` instead.

ondreian: oh, good call, I changed it from `get` to `node` to be more explicit about what we are actually getting from the cluster

ondreian:
I like the core idea of this pr, but I do not like `DCell`.  

It is not fault-tolerant, unmaintained and I don't like the security-implications of RPC for this.

I do like the idea of a redis-backed `localhost` first coordination system that can be extended to be a p2p system later, unfortunately I just don't like the idea of writing not only the localhost server, but also the eventual p2p intracluster coordination system from ground-zero in Ruby, it's just waaaay too much work for me to consider.

PR #17, opened 2018-05-18
Run spec in Travis CI

sandersch:
This updates the Travis config to run rspec as part of the build. I moved the existing `util/repo` script to the deploy step and changed the script step to run rspec. I also included a spec I had already written to make sure everything is wired up correctly.

@ondreian what is the best way to test the Travis config? I was going to turn on PR builds, but I think that would run `util/repo` and publish all in flight PRs which is no bueno.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,6 @@
 language: ruby
 rvm:
   - 2.3
-# whitelist branches that run in CI
-branches:
-  only:
-  - master
-  - canary
 
 script: true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: ruby
 rvm:
   - 2.3
 
-script: true
+script:
+  bundle exec rspec
 
 # lich repo publishing credentials
 env:
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,3 @@
+source 'http://rubygems.org'
+
+gem "rspec"
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+Gemfile.lock
--- a/scripts/rofl-puzzles.lic
+++ b/scripts/rofl-puzzles.lic
@@ -1,5 +1,5 @@
 =begin
-  version: 0.7
+  version: 0.8
   Author: Kragdruk
   Tags: festival, RoL, Rings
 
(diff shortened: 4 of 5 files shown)


.travis.yml
```diff
@@ -8,4 +8,7 @@ branches:
   - master
 
 script:
-  util/repo
\ No newline at end of file
+  bundle exec rspec
```
sandersch: I thought it made sense to invoke rspec directly instead of creating a script with solely this line. My rule of thumb for stuff like this is to create a script if it is any more complicated than a single line.

Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'http://rubygems.org'
+
+gem "rspec"
```
sandersch: I use rspec for pretty much everything but have no problem with another testing framework if you have strong opinions.

Gemfile.lock
```diff
@@ -0,0 +1,26 @@
+GEM
```
sandersch: Do we want to just ignore the lockfile for our use case?

spec/rofl-puzzles_spec.rb
```diff
@@ -0,0 +1,56 @@
+load "scripts/rofl-puzzles.lic"
```
sandersch: AFAIK there is no way to get require to load a `.lic` file. The big difference in behavior is that require will try to be smart and only execute the file once, but load will do it on every invocation. I don't think it matters for our use case, but I haven't used load for much other than this so there may be some other gotchas.

Gemfile.lock
sandersch: I thought about this a little more and I'm going to ignore it. While I would _always_ include this for a Rails app, I wouldn't in a Gem which is probably more similar to what we are doing. Our contributors likely use wildly different ruby version so locked versions will be extra annoying.

While commiting it does speed up bundle install by avoiding dependency resolution, we have a single gem in your Gemfile for now so the speed difference should be minimal.

spec/rofl-puzzles_spec.rb
ondreian: The only difference I am aware of between `load` and `require` in this context is that `load` is not cached by the global `$LOADED_FEATURES` variable

sandersch:
Travis [already skips the deployment step for PR builds](https://docs.travis-ci.com/user/deployment#Pull-Requests) so we are good to go after we separate them.

.travis.yml
```diff
@@ -1,11 +1,13 @@
 language: ruby
 rvm:
   - 2.3
-# whitelist branches that can publish
-# to the repo
-branches:
-  only:
-  - master
```
sandersch: @ondreian LMK if you think we should remove this section now, or keep it as a transitional step but remove in a follow-up PR to avoid incidental publishing of non-master changes.

PR #18, opened 2018-05-18
Move lich repo publishing to deploy step in CI

sandersch:
I was paranoid about accidentally publishing in #17, so I created this PR to just to move `util/repo` from the script step to the deploy step.


sandersch:
I'm going to merge this as I feel pretty confident it won't cause problems.

PR #20, opened 2018-05-20
Adds lich repo publishing credentials as encrypted secrets

sandersch:
Updates the Travis config to add the lich publishing credentials as encrypted secrets. While the author name isn't really a secret, I thought it made sense to stick to a single method of defining environmental variables.

After this is merged, we can remove the existing variables in the web UI. When those are deleted, we should be ready to turn on PR builds in Travis without risking accidentally publishing from existing branches.


ondreian:
I think we can probably remove [these lines](https://github.com/sandersch/scripts/blob/c9c773fb17f5177a9c44ab44c9ccd340f95eba00/.travis.yml#L4-L8) too now, right?

sandersch:
Since we have "Build pushed branches" enabled in Travis, I think we want to delete the repo credentials from the web UI before we remove those lines. I think otherwise any branches that don't include the most recent Travis config may accidentally get published.

PR #21, opened 2018-05-21
Allow all branches to build in Travis

sandersch:
Now that we have:

* Moved util/repo to the deploy step
* Set the deploy step to only run on the master branch
* Moved the elanthia-online lich repo credentials into secret
  environmental vars to prevent untrusted builds from accessing them

we can safely allow any branch to build in Travis.


sandersch:
This will be included as part of #17

PR #22, opened 2018-05-27
gameobj-data.xml specs for NPC categories

sandersch:
This adds specs for the NPC types in the GameObj XML data file. I added Lich's GameObj implementation in `lib/lich/gameobj.rb` since it doesn't seem feasible to require the full lich.rb.

I set these specs up with one assertion per example and there are roughly 3000 examples here, but the suite still completes in ~2 seconds in Travis.

While creating these specs, I found a few issues with the existing data and have left these as pending examples to fix in the future in a separate PR.

### DIFF
--- a/lib/lich/gameobj.rb
+++ b/lib/lich/gameobj.rb
@@ -1,0 +1,341 @@
+require 'rexml/document'
+
+# GameObj#load_data depends on this monkey patch
+class NilClass
+  def method_missing(*); end
+end
+
+# Tell GameObj where to find the GameObj data XML file
+class GameObj
+  DATA_DIR = Pathname.new('./scripts').expand_path
+end
+
+# Lich 4.6.44
+class GameObj
+  @@loot          = Array.new
+  @@npcs          = Array.new
+  @@npc_status    = Hash.new
+  @@pcs           = Array.new
+  @@pc_status     = Hash.new
+  @@inv           = Array.new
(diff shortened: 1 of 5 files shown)


lib/lich/gameobj.rb
```diff
@@ -0,0 +1,341 @@
+require 'rexml/document'
+
+# GameObj#load_data depends on this monkey patch
```
sandersch: As much as I dislike polluting NilClass globally with this monkey patch, it seemed better to include it than to modify the GameObj implementation below.

I've never used a [refinement](https://blog.codeship.com/ruby-refinements/) in real code, but that could be an alternative way to deal with this monkey patch without polluting NilClass everywhere.

lib/lich/gameobj.rb
```diff
@@ -0,0 +1,341 @@
...
+class NilClass
+  def method_missing(*); end
+end
+
+# Tell GameObj where to find the GameObj data XML file
+class GameObj
```
sandersch: `GameObj.load_data` does take a filename argument so this isn't strictly necessary, but I thought this was a reasonable default for this repo.

ondreian:
LGTM, eventually we'll have to migrate it to using `dist/gameobj-data.xml` but keeping the monkey-patches isolated like this should make it a breeze.

PR #28, opened 2018-07-23
add collectible as a type

mrhoribu:
Added collectible as a type.


ondreian:
The failing test is because of the new type key not existing in the tests, it does need to be fixed before we can merge this

ondreian:
This isn't going to be deployed without some whitespace tricky now, since the tests were not fixed on this branch before getting merged to master

mrhoribu:
Was afraid something like this would happen due to my rush to merge. I could change the regex pattern to swap part of it from the front to the back to force it to be re-evaluated again. But take it we would need to merge the other pull request I have open first for the core_spec?

PR #38, opened 2018-08-01
gameobj-data: Remove veil from uncommon list

sandersch:
Looks like somehow both `veil iron` and `veil` were both in the list of uncommon materials, so this PR removes `veil`. I noticed that the existing counter examples in this spec were not set up, so I fixed them and moved the gem/valuable-related assertions to the gem spec.


ondreian:
LGTM

PR #43, opened 2018-08-05
gameobj-data: clothing specs

sandersch:
Adds examples for all of the existing clothing nouns and exceptions in gameobj-data.xml.

Couple minor additions to gameobj-data.xml for consistency:

1) adds "kit" to the pawnshop sellable list. Every other clothing noun was already included in the list so this seemed like an unintentional difference
2) updates the clothing and sellable exclusions to handle "Elanthian Guilds voucher pack" the same as "Adventurer's Guild voucher pack".


ondreian:
Kits aren't common treasure i don't think

sandersch:
Removed the commit that added "kit" to the sellable list. I will make a separate issue to discuss what is in scope for sellable data.

PR #48, opened 2018-08-16
[RFC] gameobj-data: remove iron from uncommon materials list

sandersch:
I was working on excluding collectibles like "handful of iron flakes" from the uncommon list when I realized that it doesn't really make sense that iron is considered an uncommon material. Iron's not an _uncommon_ material but rather a _common_ material like "steel" or "wooden" or any other mundane metal. There are many uncommon type exclusions for iron and I have also noticed that a lot of crap critter weapons are made of iron, so I don't really think it belongs in the uncommon list.

In #38, we removed `veil` from the uncommon list. I suspect that at some point `veil iron` got separated into veil and iron as it is the only uncommon material with a space in the name.

This PR is a WIP, but before I finish it, I want to get your opinions. Do any of you see any issues with removing iron from the uncommon list?


sandersch:
From Discord:

Tillmen: I can't think of any reason iron should indicate an uncommon item.  My best guess is that someone took a list of all metals and woods and removed the common ones, and just overlooked iron.

ondreian:
I'm fine with this, but I think it need to wait until #19 is merged to master

sandersch:
Probably going to do this after #61 is merged.

PR #50, opened 2018-08-16
update collectible to include Khaarne

mrhoribu:


sandersch:
Seems legit. Can you please add an example to the [collecitble spec](https://github.com/elanthia-online/scripts/blob/master/spec/gameobj-data/collectible_spec.rb#L58)? Travis should automatically run the tests if you push an update to this branch. 

I would like to add to add some documentation for gameobj types and tests when I get a chance. I am almost done adding examples for the existing gameobj entries and would like to start on documentation and fixing data issues next.

mrhoribu:
Took me a bit, but finally found how to use the website to modify files in the same PR. So added the missing statue to the spec. Looks like it should be good.

sandersch:
:+1: Thanks for adding a test. This looks good to merge. :shipit:

PR #61, opened 2018-09-11
gameobj-data: run specs against both versions of XML data

sandersch:
Updates the Travis config to run the specs against each version of the gameobj-data.xml. Also tweaks the ENV flag to make it easier to explicitly use the non-compiled version. I added a few more specs, including some for the minor regressions I have noticed so far in the LFM traps.


type_data/tables/lm_trap.yaml
```diff
@@ -2,5 +2,5 @@ name:
   - (?:green-tinted|thick glass|clear glass) vial
   - (?:light violet|emerald|reddish-brown|wavering|icy blue|stone grey|amethyst|dark red) sphere
   - dark crystal
-  - steel needle
-  - steel jaws
\ No newline at end of file
+  - slender steel needle
+  - pair of small steel jaws
```
sandersch: @ondreian Should this be a migration instead?

ondreian: I think not, since this is in the core (previously working) set of definitions?

I want to use migrations for anything we are layering on top that currently does not exist in GameObj (like SoS, Red Forest)

sandersch: :+1: makes sense to me

PR #62, opened 2018-09-12
Create 1_old_ta_faendryl.rb

mrhoribu:
add Ithzir champion to aggressive_npcs

### DIFF
--- a/type_data/migrations/1_old_ta_faendryl.rb
+++ b/type_data/migrations/1_old_ta_faendryl.rb
@@ -1,0 +1,3 @@
+migrate :aggressive_npc do
+  insert(:name, "Ithzir champion")
+end


sandersch:
I combined both your commits into this PR and added an Ithzir champion example to the aggressive npc spec. For the time being we need to update both `scripts/gameobj-data.xml` and the compiled version, but in the long run we will only need to add a migration and test.

sandersch:
:+1: Looks good to me

type_data/migrations/1_old_ta_faendryl.rb
```diff
@@ -0,0 +1,3 @@
+migrate :aggressive_npc do
```
sandersch: @ondreian does this migration file name follow your expected naming convention?

Also should we move the other OTF npcs into here in a follow up PR?

ondreian: It does.  

For manually created migrations I want it to be `migrations/{{INDEX}}_{{CATEGORY}}.rb`

Eventually I am going to add to the CLI so that we can run `bin/migration create --name=whatever` and it create the file `migrations/{{EPOC_TIME}}_{{CATEGORY}}.rb` automagically, but when sorting the `migrations/*` folder in descending order, it should be chronological so that migrations can be properly layered without much magic.

ondreian: > Also should we move the other OTF npcs into here in a follow up PR?

I think not, I currently think it best only new changes should be added to `migrations/*`, and anything that already exists in `gameobj-data.xml` should exist immutably in `.yaml` tables

I am open to counter-arguments if you can think of any reason why this is not optimal.

sandersch: Just for attribution, i.e.since the migration is named after the area, I would expect all of the info for that area to be in there. I fine with it as-is.

PR #69, opened 2018-10-01
gameobj-data: fix issues with Red Forest skins

sandersch:
Treekin warriors yield `some blood-stained bark` as skins. This skin has a non-countable determiner (some) which is included as part of the name, unlike other skins which use the countable determiners a/an that the game presents as outside the name. These also have a slightly different bundled name (bundle of blood-stained barks) which needed to be taken into account.

The compiled version of gameobj-data.xml was also missing the "pure white feather" yielded by winged vipers.


ondreian:
This is fine to put in the yaml for now, but next time it should go in a migration, like the other red forest aggressive npcs

sandersch:
Updated this to move the skin changes into a migration.

Issue #45, opened 2018-08-12
Solhaven Herbs categorized as Skins


gs4-ashraam:
Backroom Catalog
  ----------------------------------------------------------------------
  20. a bunch of acantha leaf      30. a bunch of basal moss
  21. a bunch of wolifrew lichen   31. a bunch of pothinir grass
  22. a bunch of torban leaf       32. a bunch of sovyn clove
  23. a bunch of woth flower       33. a large bolmara potion
  24. a bunch of ambrominas leaf   34. a large rose-marrow potion
  25. a bunch of ephlox moss       35. a large brostheras potion
  26. a bunch of cactacae spine    36. a large talneo potion
  27. a bunch of calamia fruit     37. a large wingstem potion
  28. a bunch of aloeas stem       38. a large bur-clover potion
  29. a bunch of haphip root    

This categorization breaks scripts which sell items based on their type, and also the herbs are not correctly added to the useherbs script.

sandersch:
These seem reasonable to add to the herb type. I will try to get them added in the near future.

sandersch:
This should be fixed now. I just merged #70 which adds these to the herb type and excludes `bunch of cactacae spine` from skin/furrier.

ondreian:
Nice work!

PR #70, opened 2018-10-01
gameobj-data: add large quantity herb versions from Solhaven herbalist backroom

sandersch:
Addresses #45


ondreian:
This should probably be a migration

sandersch:
> This should probably be a migration

Makes sense, will update.

PR #82, opened 2019-01-02
[puritan.lic] better matching of 1004 outcome

ondreian:
the line after `improves somewhat` should be what we attempt to match


HebrewToYou:
Looks good!

PR #86, opened 2019-01-12
Orb outcome from 1004 overrides other outcomes

sandersch:
When using 1004 on a gem, the gem damage messaging (i.e. improvement or strain) comes before the messaging for a gem purified to orb status. This updates the script to continue processing the 1004 result after the gem damage messaging until sees the orb messaging, the generic non-orb messaging from 1004, or another result that cannot result in an orb gem.

### DIFF
--- a/scripts/puritan.lic
+++ b/scripts/puritan.lic
@@ -112,13 +112,13 @@ end
 ##
 module Puritan
   module Outcomes
-    ALL       = /what were you|cannot|must be holding|shatter|song misfires/i
     MAX       = /cannot be/
     SHATTERED = /shatter/
     MISFIRE   = /misfire/
     STRAIN    = /you hear it crack loudly and strain/
     PURE      = /begins to twist and turn as the very essence flows into it filling the imperfections/
     IMPROVED  = /more perfect|what were you|improves somewhat|crack|shatter|cannot be|must be holding|smoother and more pure in color/
+    ALL       = Regexp.union(MAX, SHATTERED, MISFIRE, STRAIN, PURE, IMPROVED)
   end
 end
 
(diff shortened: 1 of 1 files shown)


scripts/puritan.lic
```diff
@@ -313,28 +313,41 @@ module Puritan
...
-    waitcastrt?
-    waitrt?
-    Spell.lock_cast()
-    prep_1004()
-    fput("sing ##{gem.id}")
-    if wait_for_own_song().eql?(:ok)
-      result = matchwait(
-        #"gem becomes more perfect",
-        #"improves somewhat",
+  def self.select_outcome_from_1004_output()
```
sandersch: This logic does what I want but it feels super awkward. I don't usually have this problem in Ruby but my brain is not cooperating today.

scripts/puritan.lic
```diff
@@ -313,28 +313,41 @@ module Puritan
...
+
+  def self.outcome_of_1004(gem)
+    waitcastrt?
+    waitrt?
+    Spell.lock_cast()
+    prep_1004()
+    fput("sing ##{gem.id}")
+    if wait_for_own_song().eql?(:ok)
+      result = Puritan.select_outcome_from_1004_output()
       Spell.unlock_cast()
```
sandersch: Unrelated to this PR and I haven't noticed any issues, but it seems like we should do this whether or not `wait_for_own_song()` returns `:ok`

scripts/puritan.lic
```diff
@@ -112,13 +112,13 @@ end
...
 module Puritan
   module Outcomes
-    ALL       = /what were you|cannot|must be holding|shatter|song misfires/i
     MAX       = /cannot be/
     SHATTERED = /shatter/
     MISFIRE   = /misfire/
     STRAIN    = /you hear it crack loudly and strain/
     PURE      = /begins to twist and turn as the very essence flows into it filling the imperfections/
     IMPROVED  = /more perfect|what were you|improves somewhat|crack|shatter|cannot be|must be holding|smoother and more pure in color/
+    ALL       = Regexp.union(MAX, SHATTERED, MISFIRE, STRAIN, PURE, IMPROVED)
```
sandersch: I couldn't find anywhere the existing constant was used. I did this with reflection at first (`Puritan::Outcome.constants...`) but it felt way too cute.

scripts/puritan.lic
ondreian: consuming these pseudo `IO` objects from Lich is always awkward because the don't implement the proper `IO` interfaces.  I've written a half dozen APIs around this problem and never been happy myself yet.

ondreian:
lgtm :+1:

PR #88, opened 2019-01-13
gameobj-data: Publish generated XML and remove hand-crafted XML

sandersch:
I feel comfortable enough with the generated XML to publish it to the repo for everyone to use. This updates `bin/repo` to publish `dist/gameobj-data.xml` if there are any changes to files in `type_data/`. It also removes the old, hand-crafted XML file that is not needed anymore, removes the config option for selecting between the two files in the lich stub, and updates the Travis config to get rid of the extra build for the non-compiled XML file.


bin/repo
```diff
@@ -176,6 +176,11 @@ module Repo
 		## cast to Array from single file
 		files = [files] if files.is_a?(String)
 		
+		if files.any? { |file| file.start_with?("type_data/") }
```
sandersch: @ondreian let me know if you think it makes more sense to do this outside of the `sync` method.

type_data/tables/alchemy_product.yaml
```diff
@@ -1,12 +1,12 @@
 name:
```
sandersch: I added this change to trigger a sync of the compiled XML.

ondreian:
lgtm :tada:

PR #94, opened 2019-01-16
gameobj-data: Fix pending specs

sandersch:
Fixes all known data issues previously memorialized in the specs.

### DIFF
--- a/type_data/tables/armor.yaml
+++ b/type_data/tables/armor.yaml
@@ -4,6 +4,9 @@ name:
   - .*?chain mail
   - .*?half plate
   - .*?augmented chain
+exclude:
+  - steel shield
+  - scratched steel helm
 noun:
   - mantlet
   - doublet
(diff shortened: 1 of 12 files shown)


ondreian:
lgtm

sandersch:
Yes, I plan on starting on making ;xmlpatch obsolete next which includes adding SoS stuff.

type_data/tables/armor.yaml
```diff
@@ -4,6 +4,9 @@ name:
   - .*?chain mail
   - .*?half plate
   - .*?augmented chain
+exclude:
```
sandersch: I couldn't add these through a migration because the table didn't previously have an `exclude` key which caused my migration to fail validations. I think this is a reasonable validation to include but it might make sense to always have all three keys (name, noun, and exclude) present in the yaml files even if they have empty values.

PR #96, opened 2019-01-19
Adds create_key command for migrations

sandersch:
Adds a `create_key` command to the migration DSL. Refactors one of the existing migrations to exercise the new command.


ondreian:
lgtm, nice addition!

Issue #97, opened 2019-01-19
Migrate XML updates from ;xmlpatch


sandersch:
# Migrate XML updates from ;xmlpatch

`;xmlpatch` was created a workaround for getting changes merged into the upstream `gameobj-data.xml`. Now that we own the `gameobj-data.xml` file published in the repo, we can create migrations for the published data and make the script obsolete. I am getting started on the work and wanted to have a discussion about some of the changes required so that we're all on the same page. There are two main types of XML updates in `;xmlpatch`:

### Updates to existing types

These are pretty straightforward and I've started creating migrations for these. I should have some PRs to send soon.

* Handful of additions and exclusions to existing types (jewelry, armor, clothing, etc)
* Creatures and gems from new areas (SoS, Reim)
* Adds optional prefix `shifting` to existing box names

### New Type Categories

These are all new types that don't currently exist in `gameobj-data.xml`. I think that they are all worthwhile additions that make sense for us to merge upstream, but if anyone feels like adding these may cause issues, please let me know.

* `reim` - creature and gem category.
* `moonshard` - Duskruin quest items
* `duskruin` - area category for gems
* `pennant chase` - Pennant Chase quest items
* `noncorporeal` -  creature category
* `companion` - new NPC category for ranger companions
* `familiar` - new NPC category for wizard familiars
* `phased` - new type for phased (704) boxes

sandersch:
@ondreian @mrhoribu let me know if you see any issues with this plan especially the new type additions.

sandersch:
The only `spell-list.xml` patch is the updated 909 messaging:
```ruby
def apply_spell_patches
    # Spell data patching
    Spell[909].instance_variable_set(:@msgup, "Faint ripples in the .* form beneath you for a moment\\.")
    Spell[909].instance_variable_set(:@msgdn, "Faint ripples in the .* beneath you become apparent before quickly dissipating\\.")
end
```
which has already been updated [upstream](https://github.com/matt-lowe/Lich/blob/master/data/spell-list.xml#L897).

mrhoribu:
Btw, I added more of the RFD creatures for reim_npcs, not sure if best to put them here for you, but here was my modified reim_npcs with RFD creatures added in.

[...]

PR #103, opened 2019-01-24
Create 8_reim.rb

mrhoribu:
migration file to add normal Reim and RFD creatures and gems to gameobj-data.xml


mrhoribu:
migration to help with retiring ;xmlpatch as referenced in #97

PR #109, opened 2019-01-27
cluster: faster response processing

sandersch:
Lich's `wait_while` / `wait_until` are too slow imho for the busy wait that checks for a response to a request. This replaces the `wait_while` block in `Cluster.request` with a loop on a 1ms sleep. I tested this out and didn't notice an increase in CPU usage, but I am going to try some other things too. I think I can have `_handle_response` invoke a block or resume execution of a fiber instead of polling.

### DIFF
--- a/scripts/cluster.lic
+++ b/scripts/cluster.lic
@@ -169,20 +169,16 @@ class Cluster
         uuid:    req_id,
         channel: [payload.fetch(:channel), %(request)]
       }))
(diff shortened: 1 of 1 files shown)


ondreian:
I like the idea of a Fiber or callback block, would require me to change some scripts, but ultimately probably worth it if the performance is that bad.

ondreian:
This seems good to me, want to merge it?

PR #121, opened 2019-02-06
gameobj: Adds Duskruin gems and moonshards

sandersch:
This creates a new `event:duskruin` type populated with Duskruin gems and the moonshard quest items. This uses a new `create_table` migration DSL cmd also added as part of this PR.


lib/migration/migrator.rb
```diff
@@ -73,12 +74,28 @@ def assert_table_exists(table_name)
...
+
     def migrate(*table_names, &migration)
       table_names.each do |table_name|
         table = assert_table_exists(table_name)
         @changesets << ChangeSet.run(table, @file, &migration)
       end
       self
     end
+
+    def create_table(table_name, kind: 'type', keys:[])
```
sandersch: Looking for feedback on this API if anyone has opinions. Basic setup here is a single required positional param for the new table name, and some optional keyword params to override reasonable defaults if desired.

sandersch: Another alternative I considered that would be a little more legwork to implement but feels more DSL-ish.

```ruby
create_table(:table_name) do
  kind 'type|sellable'
  create_key(:name)

  # in theory we could also support insert/etc here but I prefer doing that in a separate migration
end
```

ondreian: I prefer the inline definitions so it's 1 method encapsulating it all, as it stands i think it's perfect the way it is.

lib/migration/table.rb
```diff
@@ -70,23 +70,35 @@ def self.validate_ruleset(table, ruleset)
...
-      @name     = File.basename(file, ".yaml").to_sym
+    def initialize(name:, kind:, rules:)
+      @name     = name.to_sym
       @log_name = %[Table(:#{@name})]
+      @rules    = rules
+      @kind     = kind
+      @pending = []
+    end
+
+    def self.from_yaml(file)
```
ondreian: perfect abstraction to keep the `new` method data store agnostic, well done.

PR #128, opened 2019-02-28
fix(wps.lic): remove registration call

ondreian:
According to Estild, wps changes are automatically registered, and this is a double registration.


sandersch:
Makes sense. Thanks for fixing.

Issue #41, opened 2018-08-04
Proposed change ot gameobj-data.xml:  Move ant larva from 'aggressive npc' to 'skin' name


OSXLich-Doug:
Ant larva listed as 'aggressive npc' type in gameobj-data.xml.

Propose moving ant larva to 'skin' name type in gameobj-data.xml.

Justification:  The game does not link ant larva (find from level 2 giant ant) in game with any tag.  Ant larva average 40 silver at furrier shops.

sandersch:
It looks like giant ants have "ant pincher" as skins. Are the ant larva loot drops that sell at the furrier like scintillating fishscales and ambergris?

OSXLich-Doug:
Yes.

Doug

sandersch:
I sent a guy a there and was able to verify this myself too. I also checked my logs and saw that the fire ants outside Ta'Vaalor also drop them. Created #42 to address.

OSXLich-Doug:
Thank you, sir!

sandersch:
Merged #42 and update pushed to the repo. These should be handled correctly now.

PR #131, opened 2019-03-17
furrier sellable fixes

ondreian:


sandersch:
That is not a comprehensive list of items that are sellable at the furrier but not exactly skins. I also know of [ant larva](https://github.com/elanthia-online/scripts/issues/41). Since you are already updating the specs, can you add ant larva and caederine to the list of examples of non-skins sellable at the furrier?

I verified that `;loot-be-gone` works fine with these changes, but I would like to better understand the consequences of this change on `;sloot sell`'s behavior before we merge this. Is there a straightforward update to `;sloot` that we can propose to work around it not selling items like this? Should be consider adding a new type similar to "valuable" for this type of item? This is a relatively small amount of items overall so its not a deal breaker for me, but I would prefer to have a reasonable solution to offer to `;sloot` users who are going to be annoyed by a breaking change triggering a known issue in widely used script.

PR #147, opened 2019-05-11
add flying check

mrhoribu:
added flying and !flying as command checks for bigshot. Useful to issue commands to knock something out of the air or to not attack if the creature is still in the air.


sandersch:
LGTM

Issue #146, opened 2019-05-11
sidewinder scale missing from gameobj-data.xml


mrhoribu:
## 🤔 Expected Behavior
Expect "a sidewinder scale" to be recognized as a skin and sellable to the furrier.

## 😯 Current Behavior
Not recognized as any TYPE data currently.

## 💁 Possible Solution
Add to both SKIN and FURRIER type data.

sandersch:
:+1: sounds good to me. Probably should be added to https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/7_sos.rb

PR #148, opened 2019-05-11
update 7_sos for sidewinder scale

mrhoribu:
fix for #146 issue adding sidewinder scale to type SKIN and FURRIER


sandersch:
Can you please add to https://github.com/elanthia-online/scripts/blob/master/spec/gameobj-data/skin_spec.rb#L195 as well?

PR #156, opened 2019-06-10
Update bigshot.lic

mrhoribu:
replaced regex match to support wands that have extra adjectives in the descriptor that could not get matched.

Example:
a silver and crimson swirled rod
in hands: a crimson swirled rod
to grab: GET CRIMSON ROD


mrhoribu:
updated to add 1020 as flee/fog option as well.

mrhoribu:
Tested for quite a few runs. Looks good. Merging.

PR #158, opened 2019-06-13
fix(gameobj): fix for bundled lich finger bones

ondreian:


type_data/migrations/23_lich_finger_bones.rb
```diff
@@ -0,0 +1,3 @@
+migrate :skin, :furrier do
```
sandersch: My preference is to name migrations like this after the area they affect (e.g. 24_rift_skins..rb) but this is better than "fix skins" or something even more generic. We haven't really arrived at conventions for these migration names so I don't feel that strongly about naming them.

ondreian: I agree, maybe even dropping the prefix number for area-specific migrations going forward.

PR #159, opened 2019-06-15
feat(cluster.lic): add Contract system

ondreian:


sandersch:
I haven't had a chance to play with this yet but seems good.

PR #168, opened 2019-08-23
gameobj: Correct invalid skin names

athias:


type_data/migrations/25_fix_undertaker_bat_skin.rb
```diff
@@ -0,0 +1,4 @@
+migrate :skin, :furrier do
```
ondreian: looks good, is there a stale rule that should be deleted due to these fixes?

athias: I attempted to remove two stale rules, however the CI pipeline failed whenever I added in the removes.
Originally I included:
`migrate :skin, :furrier do
  delete(:name, %{leathery bat wings})
  delete(:name, %{agresh bear claw})
end`
But travis CI didn't seem to like that.

mrhoribu: No further input. Closing conversation and merging PR.

PR #171, opened 2019-09-04
treim scrub update 2

HebrewToYou:
- Removing "closed" from incant 435
- Adding in a follow-up mstrike after a 435 with no rare


mrhoribu:
Looks good for ;treim, requested Ondreian's approval for puritan before I merge.

PR #176, opened 2019-09-18
Added invdb Detection

rsecor:
Added detection for invdb.lic.
If it exists then inv.db3 is also backed up.


rsecor:
I didn’t have time to think about it.  
Good call though.

PR #188, opened 2019-12-01
Update to reimmessage for officer updates

HebrewToYou:
Removing Tysong and Siierra; replacing with Nodyre and Melivn


mrhoribu:
Looks good to me

PR #224, opened 2020-04-01
EG Necrotic Ring - ghostly mobs fix

mrhoribu:
Initial start of fixing ghostly mobs to be detectable that were introduced from the EG 2019 necrotic ring.


mrhoribu:
Believe I've done this correctly. Just waiting on a review approval. Thanks.

PR #230, opened 2020-04-19
Update rofl-puzzles.lic

mrhoribu:
Added "colorful" adjective to mosaic to fix issue when having something else mosaic in your own inventory, EG a tattoo.


sandersch:
Thanks for updating this.

PR #232, opened 2020-04-21
Update rofl-questions.lic

rsecor:


rsecor:
Updated to include all currently known questions from other scripts and various sources.
This was done with the help of Soliere and a google spreadsheet that is currently shared between her and myself.

PR #233, opened 2020-04-21
Update rofl-questions.lic

rsecor:
This was done with the help of Soliere and a google spreadsheet that is currently shared between her and myself.


sandersch:
Thanks for adding these questions/answers.

PR #239, opened 2020-05-13
treim v2.20

mrhoribu:
Removed targeting of boss on boss waves as new group tag credit change should negate need to specifically hit it.


mrhoribu:
No longer needed due to #244

PR #251, opened 2020-06-01
briefcombat v0.0.8

mrhoribu:
Fixed 917 so it's identified properly instead of being lumped with 635


mrhoribu:
@tangman care to review? I've also modified my local copy to have a perm toggeable xtreme mode as I don't want to have to launch it every time with -x. So I can share that if you'd like. But basically launching it with -x would set it on until ran again with -x to turn it off.

tangman:
Looks good! 

I like the toggle idea.  Let's add an echo when they turn it on, so they know how to turn it off again.

athias:
Is this PR still valid considering it's almost a year old?

PR #253, opened 2020-06-05
feat(puritan.lic): add more config flags

ondreian:


sandersch:
Nice, new features look handy. I usually do a lot of purifying during D&D on Fridays so I will give this a shot tonight.

scripts/puritan.lic
```diff
@@ -156,8 +171,11 @@ module Puritan
       --orbs="velvet sack"              where to put orb gems
       --mage="leather case"             where to put mage rechargeable gems
       --loresing                        if we should check the purpose of a gem for sorting
-      --skip_armor                      don't remove & rewear armor
+      --skip-armor                      don't remove & rewear armor
       --silent                          skip sitrep reports
+      --min-value=<int>                 only sing to gems above a certain value
```
sandersch, line 176: Nit: Your other examples are concrete but this one is abstract and might be jargony for non-programmers. May also want to add a note that you will only do this is the character has enough training.

sandersch:
Gave this a test run tonight with the new options and it worked great for me. Thanks for the update.

PR #254, opened 2020-06-17
Added LTE Boost option

athias:
Added option to have bigshot automatically use LTE boosts if defined


mrhoribu:
Ran a few runs and looks stable to me. I always just ran ;lte to do this, but this makes it more inclusive without need for additional scripts and allows use of set # of uses per run. Nice work.

Issue #132, opened 2019-03-24
TradeBars Missing from GameObj-data.xml


mrhoribu:
a large platinum tradebar

## 🤔 Expected Behavior

To have it shown as a gem or valuable so that it can be sold via various loot scripts

## 😯 Current Behavior

Nothing as it's not classified

## 💁 Possible Solution

Add to either gems or valuables, sells for the most at the gemshop.

## 🔦 Context
Currently not picked up by loot scripts as no TYPE given for it.
Can also come in other various flavorings (gold, silver, etc). Not positive on all variations.

mrhoribu:
The tradebars can come with or without size descriptor
example:
a silver tradebar
a large platinum tradebar

Issue #129, opened 2019-03-07
waxy grey caederine 


sandersch:
## 🤔 Expected Behavior

[waxy grey caederine](https://gswiki.play.net/Waxy_grey_caederine) should be in the GameObj data.

## 😯 Current Behavior

Missing entirely.

## 💁 Possible Solution

Probably belongs in the same categories as [ambergris](https://github.com/elanthia-online/scripts/issues/106).

## 🔦 Context

Came up in Discord discussion today. Asked @HebrewToYou to save some next time he sees it in the Rift so we can test it out.

## 💻 Examples

"waxy grey caederine" is only description afaik.

sandersch:
[8:37 PM] HebrewToYou: officially confirmed: can only be sold to pawnshop or furrier. Worth 3x as much when sold to the furrier.

sandersch:
Looks like this should be:

**type:** skin
**sellable:** furrier

just like ambergris

ondreian:
I have reservations about how we are handling ambergris and caederine type items.

They should not have type `skin`, but should be sellable `furrier`, there is no way to cross reference what was something generated via the `SKIN` verb otherwise.

Loot scripts not relying on both `sellable` and `type` info are buggy, imo.

For reference this is an issue i am currently experiencing when running my selling script:
```
;lucre sell --furrier --skins=false
```
will not sell ambergris, but it should

sandersch:
I think that's a reasonable point. I would be fine with only adding caederine to the furrier sellable list and considering the removal of ambergris and similar items from the skin type.

sandersch:
In #265 I am updating these to be sellable at the furrier, but have a type of `reagent`. They are a primary ingredient in many of the dye formulas and can be stored in jars like other reagents.

PR #265, opened 2020-07-14
gameobj-data: updates for July 2020

sandersch:
Some minor updates to the gameobj data. I have been running most of these locally for a while.


sandersch:
Fixes #132
Fixes #129

PR #273, opened 2020-08-14
Bigshot: Ambush, UCS and group logic updates

wmahmoud:
# 💬 RFC

Updates ambush to not be stuck in an attack loop targeting the head and neck over and over.  Also provides some logic for creatures without a specific body part (i.e animated slush, glacial morph..etc etc)

Removed UCS mstrike target count box.  All commands were routed through cmd(), which automatically pushes all mstrikes through mstrike_cmd, so it always got priority anyway.  

Updated groups to allow leader to run scripts and perform actions without the followers spamming spells in the locksmith or gemshop from ;waggle (or any other consequences of resting scripts/actions running while walking around).  Also did some slight updates to how a hunt starts.


wmahmoud:
Didn't see the tdusk stuff get shoved in with this, closing it

PR #285, opened 2020-09-06
Added new game-object type "manna bread"

athias:
Initial creation of manna bread game object type


sandersch (approved):
:shipit: Looks good to me

PR #289, opened 2020-09-10
Uhaul patch for Kraken's Fall and Cysaegir

gs4-Xanlin:
Uhaul patch for Kraken's Fall and Cysaegir.  Removes need for trust and backs out ostruct .


ondreian:
Thanks for the contribution! 

Just for future reference semicolons are not necessary in Ruby where 1 line is 1 statement

Issue #288, opened 2020-09-10
OSA Mobs


mrhoribu:
## 🤔 Expected Behavior

OSA Mobs should be seen as aggressive

## 😯 Current Behavior

They aren't seen at all

## 💁 Possible Solution

Add to aggressive and undead appropriately 

## 💻 Examples

battle-scarred hulking krolvin huntmaster
battle-scarred hulking krolvin warden
battle-scarred krolvin huntmaster
battle-scarred krolvin huntmistress
battle-scarred krolvin warden
dwarven pillager
dwarven pirate
dwarven raider
dwarven scout
dwarven waylayer
elven brigand
elven pillager
elven pirate
elven raider
elven robber
elven scout
elven waylayer
erithian pillager
erithian raider
ethereal brigand
ethereal pillager
ethereal pirate
ethereal raider
ethereal scout
ethereal waylayer
ghostly gnomish rogue
ghostly half-elven marauder
ghostly halfling rogue
ghostly human brigand
ghostly human marauder
ghostly human mugger
ghostly human outlaw
ghostly human rogue
ghostly human thief
giantman brigand
giantman marauder
giantman mugger
giantman pillager
giantman pirate
giantman raider
giantman robber
giantman scout
gnomish brigand
gnomish mugger
gnomish pillager
gnomish pirate
gnomish raider
gnomish rogue
gnomish scout
gnomish waylayer
grizzled krolvin champion
grizzled krolvin huntmistress
grizzled krolvin warden
grizzled krolvin warlord
half-elven brigand
half-elven marauder
half-elven mugger
half-elven pillager
half-elven pirate
half-elven raider
half-elven scout
half-elven thug
half-elven waylayer
half-krolvin brigand
half-krolvin pirate
half-krolvin waylayer
halfling brigand
halfling highwayman
halfling pillager
halfling pirate
halfling raider
halfling rogue
halfling scout
halfling thug
halfling waylayer
hulking krolvin archmage
hulking krolvin champion
hulking krolvin destroyer
hulking krolvin obliterator
hulking krolvin paragon
hulking krolvin templar
hulking krolvin tormentor
hulking krolvin triggerman
hulking krolvin warlord
human bandit
human brigand
human marauder
human mugger
human outlaw
human pillager
human pirate
human raider
human rogue
human scout
human thief
human waylayer
menacing krolvin archmage
menacing krolvin destroyer
menacing krolvin obliterator
menacing krolvin paragon
menacing krolvin templar
[...]

PR #290, opened 2020-09-15
Create 33_osa_kf_npcs.rb

mrhoribu:
Initial OSA creatures for aggresive_npcs and undead types


mrhoribu:
Initial pass at adding mobs for #288

type_data/migrations/33_osa_npcs.rb
```diff
@@ -0,0 +1,104 @@
...
+  insert(:name, %{gnomish waylayer})
+  insert(:name, %{grizzled hulking krolvin huntmaster})
+  insert(:name, %{grizzled hulking krolvin triggerman})
+  insert(:name, %{grizzled hulking krolvin warden})
+  insert(:name, %{grizzled krolvin huntmaster})
+  insert(:name, %{grizzled krolvin warden})
+  insert(:name, %{half\-elven pillager})
+  insert(:name, %{half\-elven pirate})
+  insert(:name, %{half\-elven raider})
+  insert(:name, %{half\-elven scout})
```
ondreian, line 36: might make sense to have tests for these, seems weird these aren't already escaped when the regexp is compiled

mrhoribu: It might already do that, I just escaped them by default. Let me modify the code and see what happens with the built in test as other .rb migration files don't escape the characters either.

PR #292, opened 2020-09-24
fixed commas in pool party

FarFigNewGut:
INFO was updated to return commas

might have broken several scripts.


sandersch:
LGTM. Also that is a fabulous branch name.

PR #295, opened 2020-09-25
Fixed file permissions - but mostly just added healbot actions

athias:


sandersch (commented):
Seems fine. I have added a shebang and run some scripts outside of lich before, but that is an extremely uncommon case so seems fine removing the x bit from these files. Looks like these are mostly @ondreian 's scripts so he might have feelings.

ondreian:
I have no feelings. :robot:

PR #296, opened 2020-09-26
Create 34_kf_herbs.rb

mrhoribu:
add herbs from KF.


mrhoribu:
tkaro root likely can be removed as it's not really a herb and is more magical in nature to turn off now obsolete ESP.

ondreian:
lgtm, but are we removing the tkaro root?

sandersch:
imho even if the current effect of tkaro root isn't useful, it otherwise behaves like other herbs as a forageable/bundleable alchemy ingredient etc so unless it is causing pain or toil somewhere I would keep it

mrhoribu:
I'm in agreement with @sandersch and we can leave it in for now and remove it if we need to later.

PR #302, opened 2020-10-04
Fixed Poolparty deposit disk wait

daggeron:
Fixes poolparty not waiting for your disk at the locksmith pool.


mrhoribu:
Looks good to me. Can you bump the version number from 2.04 to 2.05/2.06 though as looks like the changelog got updated but not the version number. Thanks.

daggeron:
Sorry about that.  Need to create a script to bump that automatically.

PR #303, opened 2020-10-10
Fixes for tdig

HebrewToYou:
EMPTY won't grab species from the digging chests
Adding "species" to the uncollectibles array should do the trick.
Redeeming seashells after grabbing them
Fixing coin commas


sandersch (approved):
Seems good

PR #304, opened 2020-10-10
Bank teller messaging updates and fixed small timing issue

wmahmoud:
Bank teller messaging updates and fixed small timing issue for tgames.lic

### DIFF
--- a/scripts/tgames.lic
+++ b/scripts/tgames.lic
@@ -20,9 +20,11 @@
      author: Tysong (horibu on PC)
        name: tgames
        tags: games, eg, ebon gate, ebon
-    version: 1.7
+    version: 1.8
 
-    changelog:
+	changelog:
+		1.8 (2018-10-10)
+			Bank teller messaging updates and fixed small timing issue
         1.7 (2018-10-10)
             Added last two missing T5 messaging
         1.6 (2018-10-10)
@@ -333,7 +335,7 @@ def lootcheck(tier, game_location)
 			if result =~ /It is worth (.*) seashells./
 				found_seashells = $1.to_i
 				fput "redeem my seashell"
-				sleep 0.1
+				pause 0.25
 			else
 				respond "Something went wrong with seashell putting away, tell Tysong"
 			end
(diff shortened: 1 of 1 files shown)


scripts/tgames.lic
```diff
@@ -333,7 +335,7 @@ def lootcheck(tier, game_location)
 			if result =~ /It is worth (.*) seashells./
 				found_seashells = $1.to_i
 				fput "redeem my seashell"
-				sleep 0.1
+				pause 0.25
```
ondreian, line 338: I'm curious why this change was necessary?  Wouldn't it be better to verify the result versus hoping the network roundtrip times match up?
```ruby
fput "redeem my seashell"
redeem_timeout = Time.now + 5 # seconds
wait_while { Char.right.noun.eql?("seashell") && Time.now < redeem_timeout }
fail "BOOM: tell Tysong something is wrong with redeeming" if Time.now > redeem_timeout
```

wmahmoud: About 1:25 times, the original 0.1 second pause would not be enough for the game-obj to update, verified that much with a couple people. It would probably be better to just look at the hands, and I could update it to be that instead.

ondreian: I think that would probably be best, at least if we are going to change the sleep, there is no guarantee even `0.25` is the correct duration to use, so it will invariably break for someone else again.

ondreian: maybe @mrhoribu wants to put some thoughts here, this is originally his script.

ondreian:
It it intended that this pr also has changes for bigshot.lic?

wmahmoud:
That was my bad, just the tgames change was intended

scripts/tgames.lic
```diff
@@ -332,8 +334,7 @@ def lootcheck(tier, game_location)
 			)
 			if result =~ /It is worth (.*) seashells./
 				found_seashells = $1.to_i
-				fput "redeem my seashell"
-				sleep 0.1
+				dothistimeout "redeem my seashell", 3, /^You redeem your #{found_seashells} Ebon Gate seashells/
```
mrhoribu, line 337: This looks better and more feasible than previous code and gives more time for completion. Thanks.

mrhoribu (approved):

PR #310, opened 2020-10-14
Update tsquelch v2.0

mrhoribu:
Rewrote majority of all REGEX matches to be less greedy, hopefuly meaning less ReDoS lag


ondreian:
Is this stale?

mrhoribu:
Should be good with 

> Is this stale?

Not anymore, just updated it for latest version and will merge it in.

PR #314, opened 2020-10-18
feat: GTK3 - signore.lic and warnings removed

OSXLich-Doug:
Changed CONSTANT to instance variable to remove ruby warning about already initialized

Added GTK version check and enabled both GTK2 and GTK3 in single script.  Version check can probably be improved grammatically.


LeBigNasty:
Closed as OBE

PR #317, opened 2020-10-19
feat(cartograph.lic): use inline bundler;

ondreian:
WIP


ondreian:
This doesn't work, it injects a lot of stuff into the global namespace that breaks lich apis

PR #320, opened 2020-11-11
Updates to Poolparty/Oleani-lib and transfer of Asexualfavors/Milkman

daggeron:
Oleani-lib has been revamped from the ground up to help with scoping. 
All scripts that I've written now use the updated Oleani
Poolparty has been updated to address some issues that have been reported.
Asexualfavors and Milkman are being transferred to you.

Asexualfavors is the rewrite of sexualfavors that uses the resource command.
Milkman is a utility for unifying the display of LNET and ESP

### DIFF
--- a/scripts/asexualfavors.lic
+++ b/scripts/asexualfavors.lic
@@ -1,0 +1,290 @@
+#quiet
+
+=begin
+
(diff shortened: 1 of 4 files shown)


mrhoribu:
Everything I can see looks good. Merging through.

PR #321, opened 2020-11-11
Create 36_necrotic_ring_boss.rb

mrhoribu:
Added new spectral prefix for boss mobs generated from EG necrotic rings


mrhoribu:
Looks like this should be ready to go. Seems to work without hitch so far for my testing.

sandersch (approved):
LGTM. :shipit:

PR #329, opened 2020-11-29
Add cake.lic for controlling manna summons.

jasedit:
Simple script that makes it easy to choose a particular type of manna bread and stick with it.

### DIFF
--- a/scripts/cake.lic
+++ b/scripts/cake.lic
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+# Script for automatically summoning a default manna bread, or a random one depending on preference.
+#
+# Author: Ineum
+# Game: Gemstone IV
+# tags: magic, manna
+# Version: 1.0
+
+@script_name = Script.current.name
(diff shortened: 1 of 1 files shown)


scripts/cake.lic
```diff
@@ -0,0 +1,52 @@
...
+# Script for automatically summoning a default manna bread, or a random one depending on preference.
+#
+# Author: Ineum
+# Game: Gemstone IV
+# tags: magic, manna
+# Version: 1.0
+
+Vars[@script_name] ||= { manna_type: 'potato cake' }
+
+summon_type = Vars[@script_name][:manna_type]
```
ondreian, line 12: I don't think you can use `@` instance variables in an ambient context like this, won't `@script_name` always be `nil`?  Fairly sure this should be `Script.current.name`

ondreian: @sandersch thoughts?  You're the Ruby guru :slightly_smiling_face:

jasedit: Good catch - fixed!

PR #332, opened 2020-12-03
Add StormFront importing to isquelch, and accept first command

jasedit:
Adds a command to import squelches from a StormFront settings XML file. Also improves the command interface so the first command is parsed, to eliminate a separate starting command. Also adds examples and improves the help statements slightly.


mrhoribu (approved):
Everything looks good. Really like this script. Thanks again.

PR #333, opened 2020-12-04
Add export command to isquelch

jasedit:
Add a command to export squelches to isquelch, so squelches can be shared between individuals. This includes both bulk export, or specified by index.


mrhoribu (approved):
Quick work on the suggestion. Thanks!

PR #337, opened 2020-12-08
puritan: loresing to orb gems if option is enabled

sandersch:
After a recent update, ;puritan stopped saving most orb gems for me. I figured out it was checking naturally orbbed gems but was treating sung gems the same regardless of whether they were orbs or not. This updates puritan to respect the orb bag setting and also to loresing to orb gems if the loresing option is enabled to check for mage rechargeable gems.


ondreian:
thanks!

PR #340, opened 2020-12-10
Creature boon boxes

athias:
Should be good - but initial run to validate with TravisCI


sandersch:
Thanks for gathering up these description and updating the tests. We barely have any GameObj test examples with after_name descriptions so the support for them is pretty crappy.

PR #349, opened 2020-12-15
Make it more clear that jinx splits into two "commands"

slitherrr:
...with grouped subcommands. Also, --repo seems to be applicable to all
script actions, not just install and list, so I made the doc say that.


ondreian (approved):
lgtm

PR #365, opened 2020-12-19
feat(jinx.lic): add repo change command;

ondreian:


scripts/jinx.lic
```diff
@@ -563,13 +576,25 @@ module Jinx
     def self.repo_rm(repo_name)
       repo = Repo.lookup(repo_name)
       Repo.atomic {|repos|
-        fail Jinx::Error, ("repo(%s) is not known" % repo_name) unless repo[:name]
```
ondreian, line 566: This is handled by `Repo.lookup` above

scripts/jinx.lic
```diff
@@ -468,15 +468,16 @@ module Jinx
         return CLI.help()
       end
       # repo list
-      if argv.repo.eql?(true) && argv.list
+      if argv.repo.eql?(true) && argv.list.eql?(true)
```
ondreian, line 471: this ensures that these arguments are passed naked, instead of as flags, which makes the CLI more specific

example:
```
repo --list=taters # should fail, but currently doesnt
repo list               # will match this exactly going forward
```

PR #366, opened 2020-12-28
treim: Autochill when leader says KILL THY SCIRPTS

sandersch:
Extends the Cease fire command to recognize any OOC whisper from the leader that includes the phrase "KILL THY SCRIPTS". Also updates this command to break out of the loop if we leave Reim or move into the Throne Room.


mrhoribu (approved):
Updated the version in the header to match changelog, looks good to me!

PR #369, opened 2020-12-30
resource v1.0

mrhoribu:
Initial upload of release.lic


mrhoribu:
Decided to create script to merge the various resource scripts out there into one. One resource script to rule them all so to speak.

PR #372, opened 2020-12-31
Updated boon boxes to include more true known variations

athias:
As noted:
    # Boon are described by (in order):
    #   {description} {material} {noun} {after_names}

After further review of logs, this is a more accurate representation of boon boxes.
I do not know if the long descriptions are in fact required.


spec/gameobj-data/boxes_and_picking_spec.rb
```diff
@@ -46,6 +33,48 @@
       %{weathered},
     ]
 
+    # Separating Boon boxes to make it easier to read
+    # Removing boon_box_prefixes as this was found to be inaccurate after closer review
+    #boon_box_prefixes = %w[austered gilded ornate crude brass-inlaid carved delicate red cracked deeply-scored] # Similar to Box_metals -> use as desc
```
sandersch, line 38: If you want to rename this, please delete it instead of leaving it commented out.

spec/gameobj-data/boxes_and_picking_spec.rb
```diff
@@ -59,12 +88,17 @@
...
 
-    boon_boxes.each do |noun, prefix, after_name|
-      it "recognizes #{prefix} #{noun} #{after_name} as a box" do
-        box = GameObjFactory.item_from_name("#{prefix} #{noun}", noun, after_name)
+    boon_boxes.each do |noun, material, desc|
+      full_boon_box_description = "#{desc} #{material} #{noun}"
+      it "recognizes #{full_boon_box_description} as a box" do
+        box = GameObjFactory.item_from_name(full_boon_box_description)
         expect(box.type).to eq "box"
+      # Removed because long description should not need to be counted - ???
```
sandersch, line 98: We do not currently have anything in gameobj-data.xml that looks at `after_name`, but my intention with the tests is to create items that are as close as possible to what we will encounter in the game. Since you went through the work to collect the full descriptions, I think it makes sense to use the full item name in the tests.

spec/gameobj-data/boxes_and_picking_spec.rb
```diff
@@ -59,12 +88,17 @@
...
+      #it "recognizes #{desc} #{material} #{noun} as a box" do
+      #  box = GameObjFactory.item_from_name("#{prefix} #{noun}", noun, after_name)
+      #  expect(box.type).to eq "box"
       end
 
       it "recognizes shifting #{prefix} #{noun} #{after_name} as a phased box" do
```
sandersch, line 104: You can remove this example block since you moved this into the phased specs below.

sandersch (changes requested):
It looks like you have the tests updated with the new descriptions except removing the broken phased example block, but you still need to update the [migration](https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/13_use_names_not_nouns_for_loot_boxes.rb). I usually try to create new migrations, but those are kind of fiddly regexes that were easier to update in place.

If you want to get feedback on your changes more quickly than commiting and pushing to run a build, you can run the tests locally with `rspec spec/gameobj-data/boxes_and_picking_spec.rb`. You may need to `bundle install` first to get the dev dependencies installed.

Appreciate you gathering the info for these and putting the PR together. Thanks for bearing with me to get these updated. I need to create some real documentation for the gameobj-data.xml workflow but haven't gotten around to it yet.

athias:
Added Migration elements
Excluded all box variations from `uncommon`
Cleaned up code

sandersch (approved):
Looks good to me. Thanks for adding more boon boxes.

PR #373, opened 2020-12-31
Update bigshot.lic

GSIV-Nidal:
Removed 603, 611, 1603 from selfcast = true listing to make 'spellnum target' work as hunting command after the Ranger and Paladin spell list updates


mrhoribu (approved):
lgtm

PR #374, opened 2020-12-31
The selfcast check in bigshot was ugly

slitherrr:
Unnecessary cast to string, unecessary regex, massive one-liner, etc. Also includes Nidal's fix from [here](https://discord.com/channels/226045346399256576/387286877327065108/794212531518308413) but that could use a double-check for accuracy since it's actual functionality.

I haven't tested this.

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -13,7 +13,7 @@
   Setup instructions: https://gswiki.play.net/Script_Bigshot
   To help contribute: https://github.com/elanthia-online/scripts
   Full Changelog: https://gswiki.play.net/Script_Bigshot_Changelog
-  
+
   v3.93 (2020-12-31)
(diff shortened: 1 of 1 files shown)


slitherrr:
Someone's editor does not have "delete extra whitespace at end of line" turned on. To git rid of extraneous crap, use this option:

image

slitherrr:
I also changed all `if !foo A else B` to `if foo B else A`

athias:
I recommend we leave this open for the time being

There are significant changes being made, and I can incorporate this into the updated version.  Should be good to release it later today

athias (changes requested):
Please leave this PR open to have changes incorporated into upcoming v4 of bigshot. Will be followed up with closure.

athias:
Changes from this PR were rolled into PR #375 for the Bigshot 4 updates

PR #376, opened 2021-01-01
jinx: data CLI

sandersch:
Did some minor refactoring and was able to get the data list/install/update commands working. info seems like it would do the right thing if the metadata fields from the headers were there. I still need to update the usage info and implement the search command.

### DIFF
--- a/scripts/jinx.lic
+++ b/scripts/jinx.lic
@@ -164,7 +164,7 @@ module Jinx
     # dynamic reference to the $data_dir/_jinx folder
     # this will be useful for testing
     def self.jinx_data_dir()
-      File.join(data_dir(), "_jinx")
+      File.join(data_dir, "_jinx")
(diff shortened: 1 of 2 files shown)


ondreian:
Looks solid so far, please add some rspec tests whenever you can.  I'll see about cutting a new jinxp release and binary

sandersch:
Had no idea there were existing specs for jinx so amazed I didn't break any of them. Added some specs for install / update / list against a server I have running locally. Feel pretty good about these changes so going to work on releasing an updated `jinxp`. Will pick this branch back up and update the examples to use a public server when that's done.

FYI I punted entirely on Cabal support. You will get a run time error if you try to install anything. Probably worth refactoring the download/install stuff to clean up the Lich vs Cabal / Script vs Data permutations eventually.

PR #379, opened 2021-01-03
bigshot v4.1.2

mrhoribu:
update change_stance routine to not return if zealot is up due to recent spell change.


mrhoribu:
no longer needed

PR #381, opened 2021-01-05
New option to look in boxes before looting and error handling for already unlocked boxes

### LINKED ISSUE #355: poolparty - can you add a line to look in the box before dumping it?
This is almost a silly request its so basic, but I can't stand a script that picks all my boxes and dumps them in my containers without **looking to see what was in each box**. Then the only way I can tell what I got was to compare what was in my cloak before and after I ran the script. I could add a "look in box" line to ;poolparty before it dumps everything into the cloak, but it will just get overwritten with any updates. Could we add this? its also a great way to go back in logs and check what you might have gotten if you missed it.

ImperArx:
I added a new option flag to toggle looking in each box prior to looting it (--look-in-box), which is set to false by default to maintain the current behavior.  
Fixes #355 

poolparty_look_in_box_option

Additionally, added a new error and handling code for attempting to deposit a box which is already unlocked.  If this happens, the box is placed back in the loot sack and additional information is printed at the end of the deposit step.  If no such failures occur the print out at the end of the step is the same as before.

poolparty_failed_boxes_printout

### DIFF
--- a/scripts/oleani-lib.lic
+++ b/scripts/oleani-lib.lic
@@ -18,9 +18,11 @@ require "forwardable"
      author: elanthia-online
        game: Gemstone
        tags: library
-    version: 0.0.10
+    version: 0.0.11
 
    changelog:
+      0.0.11:
+        Add method for looking in boxes
       0.0.10:
         Fixed invalid method signature to quiet_command and anon_hook
       0.0.9:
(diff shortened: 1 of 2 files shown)


mrhoribu:
Did a once over, but so far so good. I did bump the required Oleani version check to match new version bump done in the PR. I'm not at a place to test it in game till later this week though, but from what I can read it looks good so far. I'll leave it open till either someone else can verify no issue or I can take a look at it later this week when I get home from travelling.

ImperArx:
@mrhoribu I've been running this for several weeks without any issues.  Is there any other testing I can provide so that we can merge it?

mrhoribu (approved):
Looks good. I'll get it merged here in a few minutes. Thanks!

PR #387, opened 2021-01-07
Updated boon boxes to remove long description until such time as they…

athias:
…'re properly recognized by GameObj

### DIFF
--- a/spec/gameobj-data/boxes_and_picking_spec.rb
+++ b/spec/gameobj-data/boxes_and_picking_spec.rb
@@ -57,6 +57,7 @@ describe GameObj do
       %{stained},
     ] # Similar to box_metals or box_woods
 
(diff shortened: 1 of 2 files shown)


spec/gameobj-data/boxes_and_picking_spec.rb
```diff
@@ -57,6 +57,7 @@
       %{stained},
     ] # Similar to box_metals or box_woods
 
+    # Leaving in descriptions in case they are used in the future
     boon_box_long = [
```
sandersch, line 61: If you remove the examples using the long descriptions, please remove this definition as well so that future readers don't think it is load bearing.

athias:
Going to merge this now and re-evaluate the long descriptions at a later time

PR #388, opened 2021-01-07
GameObj Add new types

mrhoribu:
create initial breakables gameobj table


mrhoribu:
So far this resolves the breakables and noncorporeal inclusion from xmlpatch as referenced in PR #97. Will include familiar and companion in my next few commits.

type_data/migrations/39_breakables_noncorp.rb
```diff
@@ -0,0 +1,169 @@
...
+  insert(:name, %{hollowed petrified root})
+  insert(:name, %{withered black rose})
+  insert(:name, %{hardened clump of sand})
+  insert(:name, %{large piece of sandstone})
+  insert(:name, %{handful of congealed green slime})
+  insert(:name, %{handful of congealed slime})
+  insert(:name, %{hardened dark chitin})
+end
+
+migrate "noncorporeal" do
```
sandersch, line 28: In this or a follow up you should consider going through the migrations to see if there is anything else we should add. Off the top of my head these two would definitely be good to add: https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/36_necrotic_ring_boss.rb and https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/8_reim.rb

mrhoribu: Sounds good. I went through them all but got held up on two. The AotA from EG and whether any of those creatures are non-corp: https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/35_eg_arena_of_the_abyss.rb
And if I missed anything from OSA https://github.com/elanthia-online/scripts/blob/master/type_data/migrations/33_osa_kf_npcs.rb as I'm not sure if "The Ethereal Captain" is undead as well as I don't have it marked as such. Also if "unworldly" is undead and non-corp.

ondreian:
@mrhoribu is this still applicable? I'm cleaning up stale prs

mrhoribu:
Not sure if it's still applicable. Never got around to finishing up testing but it's a good start if someone did want to tackle it.

mrhoribu:
Removing familiar and companion for now so we can get breakable and noncorporeal integrated.

PR #395, opened 2021-01-10
jinx: explicit error message for ancient ruby versions

sandersch:
Much better error message including link to wiki installation instructions for people trying to run ;jinx on ancient ruby versions. I picked 2.3 because that's when the squiggly heredoc was added which should be the most recently added ruby feature used in this script. 2.3 is already eol and 2.4 is coming up on eol in March, so I think this is a reasonable minimum version to support.

### DIFF
--- a/scripts/jinx.lic
+++ b/scripts/jinx.lic
@@ -1,6 +1,23 @@
 =begin
   Package Manager
 =end
+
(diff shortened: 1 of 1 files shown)


sandersch:
derp old ruby versions cant even parse the script so this doesnt work

PR #403, opened 2021-01-22
Create 43_legendarybox.rb

mrhoribu:
add legendary box name to existing Box table

### DIFF
--- a/spec/gameobj-data/boxes_and_picking_spec.rb
+++ b/spec/gameobj-data/boxes_and_picking_spec.rb
@@ -161,6 +161,17 @@ describe GameObj do
       end
     end
 
+    describe "legendary loot boxes" do
+      [
+        %{filigreed rolaren reliquary},
+      ].each do |box_desc|
+        it "recognizes #{box_desc} as a box" do
+        box = GameObjFactory.item_from_name(box_desc)
+          expect(box.type).to include "box"
+        end
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


mrhoribu:
Don't know of any other category to put this in other than BOX. But `filigreed rolaren reliquary` is the part that LICH recognizes and should trigger. Since currently before_name & after_name don't populate correctly.

spec/gameobj-data/boxes_and_picking_spec.rb
```diff
@@ -161,6 +161,17 @@
       end
     end
 
+    describe "legendary loot boxes" do
+      [
+        %{filigreed rolaren reliquary},
+      ].each do |box_desc|
+        it "recognizes #{box_desc} as a box" do
+        box = GameObjFactory.item_from_name(box_desc)
+          expect(box.type).to include "box"
```
sandersch, line 170: This will be "uncommon" as well because it includes rolaren in the name. I think that's fine but you can add an exclude if you think it makes sense to remove it.

type_data/migrations/39_legendary_boxes.rb
```diff
@@ -0,0 +1,4 @@
+migrate :box do
```
sandersch, line 1: FYI I renumbered this which will throw off the numbers in your other PR. This is annoying and I think we can come up with a better naming scheme to avoid this issue in the future.

PR #414, opened 2021-02-10
Update bigshot.lic

mrhoribu:
fixes cman's to be usable with the force <cmd> until <result> option.


mrhoribu:
Had Kilshaar/Sylpher test the change and seems to be working without issue.

mrhoribu (commented):
two different testers were successful in using the code and it now working properly for cman/smr results.

athias (approved):

PR #419, opened 2021-02-14
Added krolvin warmages and champion

darkshade9:


darkshade9:
I think I figured out why some of these weren't caught early on -- some of the krolvin names are specific to near-cap or capped NPCs.  For instance, I've never seen a destroyer on my own, but when I'm with near-cap or capped characters, I PEERed around and saw some I've never personally seen, so I assume the same can be said for lower-leveled krolvin not spawning for near-cap and capped.

mrhoribu (approved):
Looks good! Thanks.

PR #440, opened 2021-03-30
More OSA Krolvin

darkshade9:


mrhoribu:
Looks like our tests for Ruby 2.4 aren't working. @sandersch @ondreian any ideas on a way to fix or just remove 2.4 testing for now?

sandersch:
Looks like the action deprecated 2.4 and removed support for it. I will open a separate PR to remove it from the testing matrix. The failure is unrelated to this PR so its not a blocker for merging this.

mrhoribu:
> Looks like the action deprecated 2.4 and removed support for it. I will open a separate PR to remove it from the testing matrix. The failure is unrelated to this PR so its not a blocker for merging this.

I'm in agreement, this PR looks good so going to merge it in. Just didn't know if there was perhaps an updated action that would support it or if it's even worth doing as 2.4 is depreciated anyways.

PR #463, opened 2021-04-26
rofl-notes: Initial Creation

athias:
A short script for auto bundling notes acquired through running Rings of Lumnis


mrhoribu (approved):
I am idiot, ignore previous review. Missed the stow all. This should be good!

PR #468, opened 2021-04-27
Create invoker.lic

smileynet:
Added invoker - a script to notify players when the invoker is available and give them the option to travel there. Will automatically ask the invoker for spells if at the park.


athias:
@smileynet Updated script based on testing, and added additional error checking.

- Identified an issue with YES sometimes not being caught consistently.  It works most of the time, but will intermittently have issues.

PR #469, opened 2021-04-29
Remove ;tears reference

anthonyduer:
The error message for when the starting bonus was less than the ending bonus returned an error to the user stating to use ';tears calc [start] [end]'. This fixes the error so it says ';resource calc....', which is the correct reference since ;tears was deprecated.


mrhoribu (approved):
nice catch

PR #471, opened 2021-04-30
Add some flowers and goblets

Greminty:
@athias

### DIFF
--- a/scripts/rofl-puzzles.lic
+++ b/scripts/rofl-puzzles.lic
@@ -1249,11 +1249,14 @@ module RofL
         "Beauty"         => "orchid seedpod",
         "Betrayal"       => "wolfsbane bulb",
         "Cheerfulness"   => "crocus seed",
+        "Danger"         => "foxglove bulb",
         "Elegance"       => "begonia seed",
+        "Expectation"    => "anemone bulb",
         "Fickleness"     => "larkspur seed",
         "Forgetfulness"  => "moonflower seedpod",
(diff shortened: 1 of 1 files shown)


athias (changes requested):
Concerning the goblets - I had issues with the goblets being of the same primary color in the case of paladins/bards? (blue).  There is a secondary exception check made for that scenario.  With 'silver' being the same for clerics and empaths - this secondary exception most likely needs to be made for them as well.

Greminty:
Oh, interesting, missed that. However, I'm not sure it's the same with empaths and clerics. Here's the log snippets of the two:

> On the mahogany bar:
> Special [4]: **an elegant silver crystal goblet**, an elegant clear crystal goblet, an elegant black crystal goblet, an elegant green crystal goblet
> Total items: 4
> [rofl-puzzles]>read bar
> In the Common language, it reads:
> ~*~ I am the goblet of clerics. ~*~
> 

> On the mahogany bar:
> Special [4]: an elegant red crystal goblet, an elegant green crystal goblet, an elegant black crystal goblet, **an elegant silver crystal goblet**
> Total items: 4
> [rofl-puzzles]>read bar
> In the Common language, it reads:
> ~*~ I am the goblet of empaths. ~*~

Best I can tell, it's the same goblet for either in this case.

athias:
Sounds good!  We will find out for sure next RoL.

athias (approved):
Will validate next RoL

PR #475, opened 2021-05-05
feature(experfect.lic): add ascension support;

ondreian:


mrhoribu (approved):
Looks good. Needs the `<output class="mono"/>` fix for SF/Wiz monospaced output. But can add after.

PR #478, opened 2021-05-12
bigshot v4.3.3

mrhoribu:
Revise GTK3 check to be more compatible with older versions of GTK that don't support Gtk::Version calls

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -8,7 +8,7 @@
   contributers: SpiffyJr, Tillmen, Kalros, Hazado, Tysong, Athias, Falicor
           game: Gemstone
           tags: hunting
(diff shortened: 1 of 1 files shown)


mrhoribu:
Had someone on a really old version of Ruby install that ;version shows unknown for GTK. Believe this is just a version of GTK that doesn't support Gtk::Version calls.
```
error in Gtk.queue: uninitialized constant Gtk::Version
```
So updated to use Gtk.check_version instead. So far it works fine for the end user and myself without failing out. Should be good.

mrhoribu:
Tests from another GTK2 user with a newer version that worked on previous `;bigshot` versions also confirmed working without issue. Merging.

PR #485, opened 2021-06-03
coin4deed: Update for new EXP layout

athias:


athias:
Issue identified by Nesmeor, and fix verified by Nesmeor as well

PR #488, opened 2021-06-19
jinx: post-install hooks and mapdb.json

sandersch:
Adds post-install hooks for spell-list.xml, gameobj-data.xml, and mapdb.json. They all reload the newly downloaded data file, and the mapdb hook also attempts to download any image files referenced in the mapdb that aren't already present in the map directory.

### DIFF
--- a/scripts/jinx.lic
+++ b/scripts/jinx.lic
@@ -373,6 +373,18 @@ module Jinx
 end
 
 module Jinx
+  module Lookup
+    class << self
+      def find_asset_in(repo, name)
+        repo[:available].find do |asset|
+          URI.decode(File.basename(asset[:file])).eql?(name)
+        end
+      end
+    end
+  end
+end
+
+module Jinx
   module Metadata
     module Base
       include Enumerable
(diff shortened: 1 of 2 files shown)


sandersch:
If you want to test this before I merge https://github.com/elanthia-online/jinx/pull/45, you can 

```
;jinx repo add test https://deploy-preview-45--admiring-leakey-35fb01.netlify.app/
```

scripts/jinx.lic
```diff
@@ -372,6 +372,18 @@ module Jinx
   end
 end
 
+module Jinx
+  module Lookup
+    class << self
+      def find_asset_in(repo, name)
+        repo[:available].find do |asset|
+          URI.decode(File.basename(asset[:file])).eql?(name)
```
mrhoribu, line 380: `URI.decode` seems to be depreciated, at least in v2.7.2 of Ruby. I was able to change this to `URI.decode_www_form_component(File.basename(asset[:file])).eql?(name)` without it giving an undefined method.

scripts/jinx.lic
```diff
@@ -533,10 +543,15 @@ module Jinx
       remote_name  = Installer.normalize_filename(remote_name)
       sources = sources.map {|source| Repo.manifest(source) }
       repo = Installer.ensure_specific(remote_name, sources)
-      asset = repo[:available].find {|remote_asset|
-        File.basename(remote_asset[:file]).eql?(remote_name)
-      }
-      local_asset_name = File.basename(local_asset_name || asset[:file])
+      asset = Jinx::Lookup.find_asset_in(repo, remote_name)
+      local_asset_name = URI.decode(File.basename(local_asset_name || asset[:file]))
```
mrhoribu, line 547: Same here for depreciated URI.decode method. This should fix it: `local_asset_name = URI.decode_www_form_component(File.basename(local_asset_name || asset[:file]))`

mrhoribu (changes requested):
Looks like URI.decode has been depreciated since v2.7.2 of Ruby. Had to update to new method name in order to have it work.

scripts/jinx.lic
sandersch: Updated this to use CGI.unescape which is available in Ruby 2.1 and 2.7 and doesn't have the same issue of trying to do too much in one method that made URI.unescape obsolete.

sandersch:
> If you want to test this before I merge [elanthia-online/jinx#45](https://github.com/elanthia-online/jinx/pull/45), you can
> 
> ```
> ;jinx repo add test https://deploy-preview-45--admiring-leakey-35fb01.netlify.app/
> ```

I merged the mapdb and maps into the core repo so you should be able to 

```
;jinx data update mapdb.json --repo=core
```

with this version of ;jinx

scripts/jinx.lic
mrhoribu: Still need to update this to use CGI.escape as well. Then should be good.

sandersch: Thanks, good catch. I missed this one at first but its updated now too.

mrhoribu (approved):

mrhoribu (approved):
Looks good to go! Thanks for doing this.

ondreian:
shouldn't the map installer download a tarball from a jinx repo that away publishing can be more isolated for something like cartograph?

sandersch:
Just trying to match the existing functionality of mapdb updating since ;repo is hard down. Open to ideas on how to improve this in the future but wanted to prioritize getting something simple working.

PR #491, opened 2021-06-25
Update bigshot.lic 4.5.0

mrhoribu:
Add fog only if encumbered or injured checkbox


mrhoribu:
Tested and logic working without issue.

PR #492, opened 2021-06-27
gameobj-data: add moor eagle talons to skin and furrier

FarFigNewGut:
these show up as:
noun => talon, name => eagle talon

when held in hand, but when in a container/laying on ground as:
noun => talon, name => moor eagle talon

so need to add explicitly to be caught by sloot et. all


mrhoribu:
have a feeling that "eagle talon" being in the gameobj-data.xml was due to the wiki listing the skin as just "eagle talon" and not "moor eagle talon" so likely need to update the wiki entry as well. Not sure if there are just plain "eagle talons" out there or if it's in error for gameobj-data. Leaving it in for now for both. Thanks for the update!

PR #499, opened 2021-08-05
Update to the update-lich5 script to support 5.0.16

LeBigNasty:
Updates from Doug for post PSM3 to support 5.0.16


LeBigNasty:
Minor update to cover .14 users.

PR #500, opened 2021-08-06
Bigshot 4.6.0 - Resolve lag from bigshot_monitor

athias:


athias:
Currently staging this, and asking for a more thorough test from some others.

PR #504, opened 2021-08-11
Add option to loot boxes and put items into stow list containers

Sordal-GSIV:
This enables the loot command that appears to have been partly stubbed in the script at some point.  When enabled poolparty will first use the LOOT command to get all items that can be sorted to STOW containers and then anything left over will be emptied into the default container.


LeBigNasty:
Incorporated into new version. Closing

PR #507, opened 2021-08-17
Create newchrism.lic

LeBigNasty:
Chrism maker script


OSXLich-Doug:
Tested and approved.  Might consider naming conventions.

PR #515, opened 2021-09-03
Add Lich5 repo to jinx seed

OSXLich-Doug:
Lich5 joins the family


OSXLich-Doug:
Superseded by broader conversations - this PR is more a one-off in that context, and should not be promoted.

PR #524, opened 2021-09-10
Update lich5-update --snapshot

OSXLich-Doug:
The --snapshot feature fails in certain test cases where the existing lich version is the same as the current json lich version.  These changes address validations and permit the --snapshot feature to function.  Adding directly to master branch to limit poor update experiences.


OSXLich-Doug:
Well, halfway there - right location, wrong file. Adjusting

PR #542, opened 2021-10-26
Fixed misspelled reference

anthonyduer:
The reference to the dictionary 'material_oil' on line 754 was misspelled as 'Material_oil', causing an uninitialized variable error.


LeBigNasty:
Closed per Onedrian's comments.

PR #549, opened 2021-10-31
Replace sloot 2.0.2 with old 3.5.gtk version

LeBigNasty:
Several users are continue to report item loss from sloot 2.0.2 not honoring exclusions and saving settings. Out of an abundance of caution for new players, replacing the 2.0.2 version with the stable 3.5.gtk until major issues are resolved.


OSXLich-Doug:
Closing without commit due to conflicts.  Will restage in separate PR

PR #556, opened 2021-12-07
Update Oleani-Lib for INVENTORY FULL Change

anthonyduer:
A change to inventory full this morning broke the starting capture. The change to the regex fixes it (There may be a more regex-ish way to fix it but this has been tested to work).

### DIFF
--- a/scripts/oleani-lib.lic
+++ b/scripts/oleani-lib.lic
@@ -699,7 +699,7 @@ module Oleani
 
       lines = Oleani::Helpers.quiet_command(
           command,
-          /^(?:You are carrying nothing at this time|You are currently (?:wearing and )?carrying)/,
+          /^(?:You are carrying nothing at this time|You are currently (?:wearing)|(?:wearing and carrying))/,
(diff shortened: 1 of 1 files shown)


scripts/oleani-lib.lic
```diff
@@ -699,7 +699,7 @@ module Oleani
 
       lines = Oleani::Helpers.quiet_command(
           command,
-          /^(?:You are carrying nothing at this time|You are currently (?:wearing and )?carrying)/,
+          /^(?:You are carrying nothing at this time|You are currently (?:wearing)|(?:wearing and carrying))/,
```
FarFigNewGut, line 702: - You are carrying nothing at this time.
- You are currently wearing
- You are currently holding
- You are holding nothing at this time.

are the potential matches now it looks like for the inv full and inv hands full commands.
`^You are (?:currently )?(?:carrying|holding|wearing)(?: nothing at this time)?`

https://rubular.com/r/mkLWB9l0JE4ESr

should catch them all

PR #574, opened 2022-01-17
Add "Collectible" type support

ChickenBeard:
Added support for "Collectible" types.
Copied "bramble patch" stow logic to caederine as it was not reliably being stowed


ondreian:
.DS_Store shouldn't be committed

ondreian:
it's fine of it's in the gitignore for everyone, that would be great actually!

PR #579, opened 2022-01-29
fix bigshot animate target error

OSXLich-Doug:
logic to ignore animates in the room


OSXLich-Doug:
Trying again - without DS store

PR #581, opened 2022-02-01
Update bigshot animates, blessing and standing while loot script running

OSXLich-Doug:
Updates to BS bring rev to 4.7.4

Ignore animates as NPCs
Do not stand if looting (or loot script) is active
Update bless methods to stop if checked, hunt if not but register a bless would benefit

Concepts - Rinualdo


LeBigNasty (commented):
looks good and works well for me.

PR #588, opened 2022-02-10
Add sloot features

mikeviotti:
There are two additions to sloot:

1. Any items that can be sold at either the gemshop or the pawnshop will attempt to be sold at the gemshop
2. Added a config option for sacks to exclude from selling. The rationale here is that users might, for example, want their scrolls stored in a bag, but don't want that bag sold. Or, maybe, they don't want their overflow container(s) to be sold.

I've never done a PR from a fork before. I'm happy to resubmit if I've missed a step or five.

### DIFF
--- a/scripts/sloot.lic
+++ b/scripts/sloot.lic
@@ -144,6 +144,8 @@ setup = proc {
     gtk_locals["enable_close_sacks"].active = (settings["enable_close_sacks"] ||= false)
     gtk_locals["overflowsack"] = Gtk::Entry.new
     gtk_locals["overflowsack"].text = (settings["overflowsack"] ||= '')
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
@mikeviotti - thanks for the submission.  

I'm going to ask a couple of folks look it over, but we're very close to releasing an entirely new sloot which would replace this one.  That would mean this wouldn't live very long in this form.

Let me know if you'd like to beta-preview the newer one and see if your magic needs to be extended there.

mikeviotti:
Thanks @OSXLich-Doug. I'd be happy to beta-preview the new one. I didn't realize that was in the works, but Rin told me today. I'll close this PR.

PR #591, opened 2022-02-27
Update go2.lic

LeBigNasty:
Update for collectibles tag


LeBigNasty:
added tag for collectibles

PR #593, opened 2022-02-28
Create eherbs.lic

mrhoribu:
Initial creation of `eherbs.lic` a replacement for `useherbs.lic` with updated logic and other various fixes.


mrhoribu:
This requires the updated util.rb library that has the Lich::Util::silver_check() method in order to work properly.

PR #594, opened 2022-03-02
Added sigil of escape option

athias:


mrhoribu (approved):
This looks good to me. Didn't test ingame, but reviewing the code I don't see anything wrong.

PR #597, opened 2022-03-07
Dead stop option

ChickenBeard:
Adds an option to the Hunting tab to "Stop for dead players." It will pause bigshot when enabled and a deader is found...but I'm open to suggestions on how to handle that.


mrhoribu (approved):
Looks ok.

PR #598, opened 2022-03-07
[spell-list] fix typo for cast-proc of sigil of power

FarFigNewGut:
the dothistimeout in the cast proc is wrong.

old: "As you concentrate on **the signal**"
correct: "As you concentrate on **your sigil**"


mrhoribu (approved):
Looks good.

PR #603, opened 2022-03-08
Rebase with newer EO/Lich-5

LeBigNasty:
Also update this repo to 1.30


mrhoribu:
Will this work on 5.3.0 without the realID lich changes? Do we need to wait for 5.4.0 to be "released" before merging? @LeBigNasty

mrhoribu:
Confirmed will work without issue. Should be good to go.

PR #610, opened 2022-03-09
[spell-list] quick fix for spell-list to not break waggle

FarFigNewGut:
should fix this proper but, just to get waggle working for people while we figure out what to do.


FarFigNewGut:
FINE THEN

PR #634, opened 2022-03-23
[script][go2] - Bring in DR updates including DR map stringproc overrides

asechrest:
Update to version 1.32. Version 1.31 brings in DR changes. Version 1.32 brings in yaml-based map stringproc overrides for DR, which provide a structure for fixing Lich5 travel hiccups and for power users to set up their own stringprocs.

### DIFF
--- a/scripts/go2.lic
+++ b/scripts/go2.lic
@@ -9,10 +9,14 @@
    original author: Shaelun
               game: any
               tags: core, movement
-           version: 1.30
+           version: 1.32
           required: Lich >= 4.6.14
 
    changelog:
+      1.32 (2022-03-22):
+        For DR: add support for yaml-based map stringproc overrides
+      1.31 (2022-03-10):
+        Merged in DR changes from v1.22f by Sarvatt
       1.30 (2022-03-03):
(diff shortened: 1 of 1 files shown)


scripts/go2.lic
```diff
@@ -161,6 +164,30 @@ show_help = proc {
...
+# For Dragonrealms only
+# A change in behavior from Lich4 to Lich5 caused certain ;go2 map wayto movements to fail.
+# This is because of the way the map stringprocs are structured, and the method used to evaluate
+# certain map moves, which involved an arguably dubious overloading of FalseClass on Lich4.
+# The transition to Lich5 makes it necessary to fix some stringprocs as an interim solution, at
+# least until the map can be divested from current owner control and handled appropriately. This code
+# also allows the flexibility to pull custom map wayto overrides from the standard yaml hierarchy.
+if XMLData.game =~ /^DR/
+  # Get yaml settings
+  settings = get_settings
```
mrhoribu, line 176: In case someone doesn't have dependency DR running, should this all be wrapped in `if defined?(get_settings)` as to avoid breaking someone that has yet to install dependency?

PR #637, opened 2022-03-25
stance perfection, assume aspect, briar weapons support and dhurl fix.

snperry85:
cman stance perfection - use 10,20,30,ect in attack stance box to use stance perfection if you know it.
assume aspect - use '650 lion wolf' '650 lion evoke' in society box or 'assume lion wolf' 'assume lion evoke' in hunting commands.
briar weapons - 'briar \<weapon\>' 'briar bow' 'briarbow' in hunting commands to activate +AS buff.

dhurl fix - added weapon bond return message to monitor so dhurl/recover can see if a weapon auto returns from bonding and doesn't end up in a recover loop. Warriors rejoice.


snperry85:
that's 'briar <weapon>' ... 'briar' by itself will do nothing.

PR #668, opened 2022-04-05
[eloot.lic] push for jinx refresh

mrhoribu:
push change for ;jinx update


mrhoribu:
Jinx finally caught up and good.

PR #669, opened 2022-04-06
[workflow] publish eo-lich5

mrhoribu:
Publish gameobj-data.xml, spell-list.xml and various base .lic scripts to EO-Lich5


OSXLich-Doug:
Exciting.  Why exclude version.lic and xnarost.lic, though?

Issue #656, opened 2022-04-04
[gameobj] Add missing hide bundles


LeBigNasty:
## 💻 Examples

triton spines, claws, hides and firebird feathers are not present in gameobj

mrhoribu:
We'll need examples of each of these things to try to properly go about testing and adding

NPC Gathered from: 
Unbundled full name: 
Bundled full name: 

Thanks.

mrhoribu:
Looking through the existing gameobj-data.xml I see the following for individual usage of the four you gave. I can't find what `claw` you may be referring too so will need more info on that. Also need to get a sample of these and what they fully look like bundled.

- elongated triton spine (triton radical, nelemar)
- iridescent triton hide (triton magus, nelemar)
- soft red firebird feather (firebirds? wiki has it as just a red firebird feather, F'Eyrie)

Will need to get on Test instsance and go grab some of these and see what can be done.

LeBigNasty:
bundle of triton spines
bundle of triton hides
bundle of red firebird feathers
bundle of gold-flecked claws

mrhoribu:
Should now be resolved.

PR #672, opened 2022-04-06
[gameobj] missing skin/furrier bundles

mrhoribu:
Add missing bundles for #656

### DIFF
--- a/spec/gameobj-data/skin_spec.rb
+++ b/spec/gameobj-data/skin_spec.rb
@@ -46,7 +46,6 @@ describe GameObj do
         %{crocodile snout},
         %{crooked crone finger},
         %{crooked witch nose},
-        %{curved gold-flecked claw},
         %{cyclops eye},
         %{daggerbeak wing},
         %{dark panther pelt},
@@ -56,7 +55,6 @@ describe GameObj do
         %{dirge skin},
         %{dobrem snout},
         %{eagle talon},
-        %{elongated triton spine},
         %{faceted crystal crab shell},
         %{faeroth fang},
         %{faintly glowing worm skin},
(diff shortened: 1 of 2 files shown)


ondreian:
I would probably add some unit tests for these, just to be sure

spec/gameobj-data/skin_spec.rb
```diff
@@ -324,6 +321,60 @@
...
     end
 
+    describe "bundle skin oddities - recognize single skin, but not bundle" do
+      [
+        %{elongated triton spine},
+        %{iridescent triton hide},
+        %{soft red firebird feather},
+        %{curved gold-flecked claw},
+      ].each do |skin_name|
+        it "doesn't recognizes single #{skin_name} as a skin" do
```
ondreian, line 331: shouldn't this be does recognize?

spec/gameobj-data/skin_spec.rb
```diff
@@ -324,6 +321,60 @@
...
+          expect(skin.sellable.to_s).to eq "furrier"
+        end
+
+        pluralized_skin_name = if skin_name.end_with? "s"
+                                skin_name
+                              else
+                                "#{skin_name}s"
+                              end
+
+        it "recognizes bundle of #{pluralized_skin_name} as a skin" do
```
ondreian, line 343: shouldn't this be doesn't?

ondreian (changes requested):
i think you have 2 statements inverted, but otherwise looks good now

spec/gameobj-data/skin_spec.rb
mrhoribu: Yep, forgot to update this when copied again to do reverse lookup.

spec/gameobj-data/skin_spec.rb
mrhoribu: Yep, forgot to update this when copied again to do reverse lookup.

PR #675, opened 2022-04-07
[ewaggle] change name add bail option

mrhoribu:
Add bail option to stop casting if out of resources.


mrhoribu:
Redoing as PR #1131

PR #678, opened 2022-04-08
[jinx.lic] prevent download as opposite type

mrhoribu:
Prevent downloading of data as a script to script folder and vice versa.


ondreian (commented):
LGTM, but probably should add a test to the jinx unit tests and try to download a script as a data, just to be sure.

PR #690, opened 2022-04-15
[bigshot.lic] swift justice tracking and command check, hide attempt change

snperry85:
will track swift justice charges. (justice) do skill if have charges (!justice) do skill if don't have charges
Tysong's adjustment to hide command to specify attempts to try.


mrhoribu (approved):
Looks good.

PR #691, opened 2022-04-15
[eloot.lic] weapon displayer support

Deysh:
v1.2.6 (2022-04-15)
    - support for weapon displayers


mrhoribu (approved):
Sent to a tester and working good now for weapon displayers.

PR #712, opened 2022-05-07
[eloot.lic] v1.3.0 bugfixes and exclusion loot/skin

Deysh:
eloot updates


mrhoribu:
Looks good, need to push due to _drag error reported via GMs. Will push out bug fixes if any found after.

Issue #673, opened 2022-04-06
[go2] Kill go2 when get_silvers/bank do not have enough money


LeBigNasty:
| Software         | Info       |
| ---------------- | ---------- |
| ruby version     |
| script           |

# 🐛 bug report

## 🤔 Expected Behavior
Go2 should die at the bank when it goes to witthdraw money and the user doesn't have enough. Presently it just continues to the point where money is needed and errors but doesn't die. 
An enterprising person could add getting silvers from FWI as a flag/option if the local bank doesn't have funds. 

The teller flips through the books and then looks up with an apologetic expression and says, "I'm sorry, Rinualdo, you don't seem to have that much in the account. 
[go2]>ask portmaster about travel 5
[go2]>ask portmaster about travel 5
Dylid says, "Well, looks like I got a captain willing to take you to Kraken's Fall.  I can't vouch for him, but you'll probably make it there dead or alive.  It will cost about 25000 silvers.  Let me know if you're ready!"
[Repeating this command within the next 30 seconds will confirm your command.]
>
Dylid growls before saying, "If you aint got the money, you aint gettin the honey!"

PR #724, opened 2022-05-12
[go2.lic] v1.35 exit if not enough silvers

mrhoribu:
Fix for using go2 with getsilvers enabled and not having enough at bank. Will exit with an echo statement stating such. Addresses #673


mrhoribu:
closes #673

PR #753, opened 2022-06-14
[gameobj] disk exclusion uncommon

mrhoribu:


mrhoribu:
@ondreian @sandersch Haven't used regex in an actual migration file before. Will this push over properly if merged?

ondreian:
it should, you probably also need `coffin` for Halloween when all disks turn to coffins

PR #756, opened 2022-06-15
[xnarost.lic] Update @theme_state

OSXLich-Doug:
Somewhere we lost the @theme_state rewrite shifting from 'on | off' to 'true | false'.  This change corrects @theme_state to be 'true | false' as it is presently stored / used in lich5.

### DIFF
--- a/scripts/xnarost.lic
+++ b/scripts/xnarost.lic
@@ -17,10 +17,13 @@ Tracks your current room on Tsoran (or other) maps
 			game: Gemstone
 			tags: core, movement, mapping
 			required: Lich > 5.0.1
-			version: 1.0
+			version: 1.0.1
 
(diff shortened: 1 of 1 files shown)


scripts/xnarost.lic
```diff
@@ -265,10 +265,10 @@ map_offset_y = nil
 narost_exit = false
 dragging = false
 
-if @theme_state == 'on' && File.exist?('maps-dark')
+if @theme_state && File.exist?('maps-dark')
```
mrhoribu, line 268: Should probably also account for older versions of Lich that still use on/off values. Otherwise could cause users who update to this version of xnarost but are still on old Lich to no longer be able to use dark-mode maps.

PR #762, opened 2022-06-30
[spell-list.xml] update wither 1115

jscholly:
adding channel = 'yes' to the spell wither (1115)

### DIFF
--- a/scripts/spell-list.xml
+++ b/scripts/spell-list.xml
@@ -1196,7 +1196,7 @@ Changelog
    <spell availability='all' name='Organ Scar Repair' number='1114' type='utility'>
       <cost type='mana'>if (Scars.torso &gt; 1) and (Spells.empath &gt;= 18) then 18 else 14 end</cost>
    </spell>
-   <spell availability='all' name='Wither' number='1115' type='attack'>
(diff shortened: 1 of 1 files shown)


ondreian:
i don't think this is the correct thing to do, this should be handled in a way that doesn't introduce a breaking change this every user

@mrhoribu thoughts?

mrhoribu:
@ondreian Correct, this would default `Spell.cast` to always channel this spell. Causing a lot of confusion. 

@jscholly if you want to use `Spell.cast` with channeling, you can do `Spell[1115].channel = true` in your script to then do a channel by default. But making a hard RT for wither default for everyone would be an issue. You can also use `Spell.force_channel` now as well to always try to channel a spell.

> adding channel = 'yes' to the spell wither (1115)

jscholly:
My apologies, I wasn't thinking about unintended consequences for others,
just accuracy in the database.  Bone shatter and most other spells (that I
spot checked) that can be channeled include the tag.

I can make a work-around in my script.  I noticed it because I'm using
Spell.channel in one of my scripts to return true or false and it behaves
differently using 1115 than 1106.

PR #840, opened 2022-10-07
[repository.lic] v2.40 remove Lich update function

mrhoribu:
remove lich update function ability and point to `;lich5-update` instead.


LeBigNasty:
should we also remove/comment out 
 output.concat "   set-lich-updatable        download updates for Lich (download-updates)\n"
output.concat "   download-lich             download the latest version of Lich\n"

PR #864, opened 2022-10-30
[bigshot.lic] v4.13.0 GUI updates and more

Deysh:
Added profile support to UI  
Added standard Lich messaging functionality
consolidated some code


snperry85:
command_check() does not support multiple checks. The final one to be checked is the only one that matters.

[Bigshot: cmd force 611 until 101(frozen m11)]
[Bigshot: command_check force 611 until 101(frozen m11)]
[Bigshot: command_check frozen 0]  <- indicates the target is frozen and should skip 611
[Bigshot: command_check m11 false] <- indicates I do have enough mana
[Bigshot: command_check final decision false] <- returns false (continue with command) because mana check over wrote frozen check.

scripts/bigshot.lic
```diff
@@ -1300,7 +1295,8 @@ class Bigshot
             'weed'    => !( Effects::Buffs.time_left("Tangleweed Vigor") <= (amount / 60.to_f) ),
           }
           
-          should_return = buff_check[command]    
+          echo "command_check section 2 #{s} #{split_check[split_item]}" if $bigshot_debug
+          return split_check[split_item] if split_check[split_item] != false
```
snperry85, line 1299: echo "command_check section 2 #{s} #{buff_check}" if $bigshot_debug
return buff_check if buff_check != false

scripts/bigshot.lic
```diff
@@ -1369,17 +1365,15 @@ class Bigshot
             end
             return false           
           else     
-            should_return = other_checks[item] 
+            echo "command_check section 3 #{s} #{split_check[split_item]}" if $bigshot_debug
+            return split_check[split_item] if split_check[split_item] != false
```
snperry85, line 1369: echo "command_check section 3 #{s} #{other_checks}" if $bigshot_debug
return other_checks[item] if other_checks[item] != false

Deysh: Thanks, got them.

PR #870, opened 2022-11-03
Added suffused, weekly and total resource values to the 'all' output

tparkhill5:
I thought it would be helpful to be able to keep track of characters available resource and suffusion values


mrhoribu:
Added suffused, weekly and total resource values to the 'all' output

PR #880, opened 2022-11-09
[ebounty.lic] initial creation

Deysh:


scripts/ebounty.lic
```diff
@@ -997,6 +1020,29 @@ module EBounty
...
     
+    end
+  
+    def self.set_eval
+    
+      bounty_eval = "(checkbounty =~ /^You (?:succeeded|have located)|your task is failed/)" # bounty complete
+      
+      if checkbounty =~ /A local divinist has had visions of the child fleeing/
+        bounty_eval += " || (checkbounty =~ /^You have made contact with the child/ && GameObj.npcs.any? { |npc| npc.name =~ /\\bchild\\b/ })"
+      elsif checkbounty =~ /^You are not currently assigned / && !Spell[9003].active?
```
snperry85, line 1031: && !Spell[9003].active?
This needs to be removed. It doesn't set the evaluation, because the only time you go out and hunt when you are not on a current bounty is if Spell[9003] is active, otherwise you're grabbing a bounty. So this condition is never met. This then causes bigshot to not have an eval that is possible, so it hunts until you run out of resources or get encumbered instead of heading back in to grab a bounty when it's available.

scripts/ebounty.lic
```diff
@@ -997,6 +1020,29 @@ module EBounty
...
+        bounty_eval += " || (checkbounty =~ /^You are not currently assigned / && !Spell[9003].active?)"
+      elsif checkbounty =~ /You have been tasked to retrieve(.*)of at least/
+        EBounty.data.bundle_skin = $2 if checkbounty =~ /You have been tasked to retrieve (\d+) (.*?)(?:s)? of at/        
+        EBounty.data.bundle_skin = EBounty.data.bundle_skin.match(/(\w+)\s+(\w+)\s*\Z/)
+   
+        bounty_eval += ' || (count = 0;GameObj.inv.each{|i| i.contents.each{ |k| if k.name =~ /bundle of #{EBounty.data.bundle_skin}/ && !hidden?; lines = Lich::Util.quiet_command_xml("measure #{k}", /You glance through/);count = $1.to_i if lines.any?{ |l| l=~ /count a total of (\d+) <a exist/};end;};};count >= EBounty.data.remaining_skins.to_i)'
```
snperry85, line 1037: I'm pretty sure this needs to be `count = $1.to_i if lines.any?{ |l| l=~ /count a total of (\d+) <a exist/}` Otherwise your count will only be of the last bundle you measured and bounties are often more than 9 skins.

PR #886, opened 2022-11-14
[gameobj] rare treasure

mrhoribu:
Add small totem as a magic item and sellable at pawnshop


mrhoribu:
fixes and closes #809

PR #890, opened 2022-11-22
[scripts][go2] Adjusting for features that exist (and don't) in DR

MahtraDR:
Moved some setting definitions around so DR would get the ones that exist there, and gating ones that don't exist behind a game check.

Also brought the custom wayto functions up to date with the current dr-scripts version.


asechrest:
@MahtraDR My `;go2 goback` PR on our side is now complete, so if you want to grab those changes and add them here.

PR #906, opened 2022-11-30
chore(ci): add syntax verifier to Github actions;

ondreian:


mrhoribu:
Using this as original inspiration, we were able to implement `ruby -c` via #1042

Issue #912, opened 2022-12-01
[gameobj] rosespar issue


gs4lass:
### Brief Description

Please add "a piece of rosespar" as a gem. It is a cheap gem, as seen here https://gswiki.play.net/Rosespar, and likely just got overlooked

### Expected Behavior

As a gem

### Current Behavior

You remove a piece of rosespar from in your sturdy leather pack.
>;e echo GameObj.right_hand.name
--- Lich: exec1 active.
[exec1: rosespar]
--- Lich: exec1 has exited.
>;e echo GameObj.right_hand.type
--- Lich: exec1 active.
[exec1: ]
--- Lich: exec1 has exited.

### Possible Solution

Adding it to game obj data or other appropriate place.

mrhoribu:
FYI items when in your hand are going to show all sorts of odd types due to how the game changes how they are shown and displayed. You can test for type on an item when it's on the ground or in a container. If you have `;foreach` installed just do a `;foreach in container` and it should show you what's in the container and what types are associated with said items.

`piece of rosespar` is actually in GameObj-data.xml already but it's listed as a `valuable` and as sellable to the `gemshop`. Unless the piece of rosespar can be purified/sung to as a gem, it should remain as type valuable and not changed to gem.

gs4lass:
You glance down to see a piece of rosespar in your right hand and nothing
in your left hand.
>1004 rosespar
[spellcaster]>prepare 1004
[spellcaster]>cast rosespar
>
You begin your musical chant for Purification Song...
Your spellsong is ready.
>
You sing a melody, directing the sound of your voice at a piece of rosespar.
In response to your voice, the rosespar glows luminously, its crystalline
structure reforming as the gem becomes more perfect.
When the song ends, you notice the rosespar sparkles a little more and is
void of some natural imperfections.
Sing Roundtime 3 Seconds.

As you sing, you feel a faint resonating vibration from the rosespar in
your hand, and you learn something about it...

You sense the item has some sort of magic resistance.  From the pitch of
the vibration you determine that the purpose of the rosespar is as a gem of
some kind.
Roundtime: 10 sec.

mrhoribu:
So it can be purified. I’ll add it to the list to change it from valuable to gem then. Thanks for confirming it’s an actual gem by attempting to purify it.

gs4lass:
Glad to help.

If that's the criteria then you might want to look into switching a piece
of iridescent mother-of-pearl from gem to valuable because it neither sings
[...]

PR #920, opened 2022-12-05
[gameobj] piece of rosespar

mrhoribu:
Change piece of rosespar from `valuable` to `gem`


mrhoribu:
Fixes and closes #912

PR #923, opened 2022-12-05
[Spell-list.xml] Update 620 to be stackable

snperry85:
Updated 620 from refreshable to stackable. Sorry for the extra PR. Should have caught it needed to be changed the first time.


mrhoribu:
This may not be valid as all research shows that this spell most likely is supposed to be refreshable. So likely a bug that it's currently stackable. Waiting on confirmation via Discord's official #Rangers channel

PR #958, opened 2023-01-10
[spell-list.xml] Add Telekinetic Disarm

mrhoribu:


mrhoribu:
Not currently released.

PR #998, opened 2023-01-25
[ebounty.lic] v1.0.31

Deysh:
v1.0.31 (2023-01-24)
    - reordered casting of song of peace during foraging and stopping it when done.


mrhoribu (approved):
Looks good and prevents recasting more than once.

PR #999, opened 2023-01-25
[bigshot.lic] v4.14.0

mrhoribu:
add new slayer command for 240


snperry85:
https://github.com/elanthia-online/scripts/blob/2fa1beaf409dead6258ff6fb4be2af095a6b403f/scripts/bigshot.lic#L4164
Let's make this  `pause @WANDER_WAIT unless should_flee?` to prevent standing there for things like clouds.

PR #1005, opened 2023-01-27
Update ebounty.lic

ServerFailure:
Fixing 1011 during foraging


LeBigNasty:
Unfortunately this won't resolve the issue. Should be resolved in #1006

PR #1016, opened 2023-01-31
[bigshot.lic] updates to room claim

snperry85:
Entering a room pulls a buffer, buffer is checked for Also here: indicating another player was there first and claiming the room. If this returns nothing it then checks against checkpcs to verify incase buffer was over run by behind the scenes commands. Removes players if they have a claim and walk out of the room. So having a wait before moving time can be beneficial when you end up "chasing" someone.

### DIFF
--- a/bigshot.lic
+++ b/bigshot.lic
@@ -1,0 +1,5561 @@
+=begin
+
+  bigshot.lic: the new optimus prime
+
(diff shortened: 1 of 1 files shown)


snperry85:
I'm really failing at this PR thing.

PR #1030, opened 2023-02-07
[bigshot.lic] v4.14.6 fixes: head/tail, cast_signs, mstrike lockout

snperry85:
Followers don't need to be told if the room is claimed, the head tells them its ok to attack. Removed the follower stuff from the claim system. cleaned up and simplified the buffer check.
added mstrike lockout message to assault regex.


mrhoribu:
@snperry85 Don't forget to bump the version number.

PR #1033, opened 2023-02-10
Update go2 with various tags

LeBigNasty:
Added tags for treasuremaster, portmaster, cobbling, and grocer


mrhoribu (approved):
Looks good.

PR #1037, opened 2023-02-13
[ecleanse.lic] v1.0.9 disarm log & casting check improvement

Deysh:


mrhoribu:
added casting check and improved some of the logic in disarm recovery

PR #1040, opened 2023-02-17
[eherbs.lic] First pass at supporting survival kits. will heal from kit, no stocking yet

Deysh:


LeBigNasty:
Stocking was added and a change in bundle method

PR #1041, opened 2023-02-19
chore: ruby syntax standalone check

mrhoribu:
Add a github workflow to check ruby syntax using the built in `ruby -c` option.


MahtraDR:
Seeing this made me go look at how to filter and only check syntax of ruby files... I've updated the ones in out repo
https://github.com/rpherbig/dr-scripts/blob/master/.github/workflows/ruby_syntax_checker.yml
and
https://github.com/rpherbig/dr-scripts/blob/master/.github/workflows/rubocop.yml

PR #1047, opened 2023-02-20
Bug fixes

tparkhill5:
- Fixed reference to @skip_scars
- Updated Utility.set_preferences to save to EHerbs.data rather than out-of-scope instance variable


LeBigNasty:
Thanks for the feedback. This was incorportaed in #1048 with a few other fixes.

PR #1052, opened 2023-02-20
[version.lic] v0.57

mrhoribu:
Syntax & Style updates for Rubocop compliance


mrhoribu:
Will revisit once Ruby 3.2 is required with newer Lich.

PR #1053, opened 2023-02-20
[eherbs.lic] v1.5.4 update

gs4-Xanlin:
typo fix, old compat fixes


mikeviotti (commented):
Typo (Tonic instead of tonis) on line 339 - not part of this PR but maybe it can be changed since you've grabbed a few other typos

Issue #1074, opened 2023-02-26
Eherbs does not work if your herbsack noun is the same as another container in your inventory and it appears first in your inventory[script] 


JaredDiCioccio:
### Script Name

eherbs.lic

### Bug Report

Assuming you work around #1069 , Eherbs still breaks after successfully saving your herbsack. 

If you have another container on you with the same noun as your herbsack, Eherbs will not successfully find and set your herbsack when it tries to use herbs. 

This behavior is entirely dependent on the order your inventory is sorted in. 

These are the containers on my belt:

**`a punched suede sack drawn by feathered ties, a studded black suede belt pouch with a silver-edged flap, an indigo elesine sack embroidered with coils of shadowy eels,`**

The suede sack is my survival kit, the elesine sack is just a container.  In this configuration, Eherbs.lic will find my suede sack successfully when calling Utility.find_herbsack. 

However if the elesine sack appears first, Utility.find_herbsack will find it first and use it as the herbsack. It will then think you have no herbs. This breaks basically all expected behavior of Eherbs. 

### Expected Behavior

I expect to run ;eherbs and have it find my survival kit and then use doses from it to heal me. 

Also, ;eherbs stock breaks too since it keeps buying herbs and then stowing them in your herb stow, but then never seeing that you have anything in your herb stow.

### Current Behavior

It looks in the wrong container so thinks you have no herbs. 

### Possible Solution

The Utility.find_herbsack function is broken. 

This code block is used to find the GameObject for your herb container. 

   ```
   unless (
        herbsack = (
          GameObj.inv.find { |obj| obj.noun == EHerbs.data[:herb_container] } || 
          GameObj.inv.find { |obj| obj.name == EHerbs.data[:herb_container] } || 
          GameObj.inv.find { |obj| obj.name =~ /\b#{Regexp.escape(EHerbs.data[:herb_container])}$/i } || 
          GameObj.inv.find { |obj| obj.name =~ /\b#{EHerbs.data[:herb_container].split(' ').collect { |n| Regexp.escape(n) }.join(".*\\b")}/i }
        )
      )
```

In the broken scenario here, my herb_container is set to `punched suede sack`. However since i have an `elesine sack` sorted before my `punched suede sack` on my belt, this block finds the elesine sack first. 

It then thinks my elesine sack is my herbsack. Since it is not, Eherbs finds no herbs in it and will always go to buy new herbs regardless if your actual herbsack is full. 

[...]

Issue #1069, opened 2023-02-24
Eherbs set herbsack command is broken[script] 


JaredDiCioccio:
### Script Name

eherbs.lic

### Bug Report

When runnng `eherbs set herbsack <your herbsack>`  the herbsack does not get set because of the parameters passed to the Utility.settings_herbsack function. 

I'm not sure if setting herbsack from the gui works. It doesn't appear to. 

### Expected Behavior

My herbsack should get set. 

### Current Behavior

The Utility.settings_herbsack function exits prematurely at https://github.com/elanthia-online/scripts/blob/master/scripts/eherbs.lic#L2298

### Possible Solution

https://github.com/elanthia-online/scripts/blob/master/scripts/eherbs.lic#L2500

The CLI is passing `Script.current.vars[0]`, which is a string. So `return unless vars[2] == 'herbsack'` will of course return because it's just checking vars[2] which is a string, so it'll always just be the letter t. 

### Possible Solution

Should just pass Script.current.vars to the function.

Deysh:
Thanks for the fix on the CLI. It should work on the next release. 

As for the noun issue, if you have multiple containers with the same noun you'll need to be more descriptive in the herbsack setting. You just need to enclose it with parenthesis - ;eherbs set herbsack "blue sack"

PR #1075, opened 2023-02-26
[eherbs.lic] fixed the 'in,on,under,behind' options, added some additional inventory handling

Deysh:


mrhoribu:
Should address #1074 and #1069

PR #1100, opened 2023-03-12
chore: infomon.rb library scripts compatibility

mrhoribu:
Update various EO\scripts to not allow running of infomon.lic script when on newer version of Lich5 that has infomon as a .rb library.


LeBigNasty:
Is this core still needed?

PR #1108, opened 2023-03-17
[eherbs.lic] v1.5.12 bugfix's for sigil of mending, leaving people at bank, closing containers. Added combined option for stocking

Deysh:


mrhoribu (approved):
Looks good now. Thanks for the header fix.

PR #1119, opened 2023-03-28
[elogin.lic] v1.1.2 CLI login change

mrhoribu:
Change from using SAL login method to native character saved entry CLI login method

### DIFF
--- a/scripts/elogin.lic
+++ b/scripts/elogin.lic
@@ -15,8 +15,10 @@
                game: Gemstone
                tags: login
            required: Lich > 5.0.1
-            version: 1.1.1
+            version: 1.1.2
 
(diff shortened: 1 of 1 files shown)


mrhoribu:
Update also provides compatibility with elanthia-online/lich-5#346

OSXLich-Doug:
SAL is required to launch SIMU FE.

gs4-Xanlin (approved):

gs4-Xanlin:
> SAL is required to launch SIMU FE.

This change skips generating the SAL with one command (in the script) and sending it with another and lets lich.rbw handle generating the SAL internally.  Tested and still works launching both wrayth and wizard fe.

PR #1127, opened 2023-04-05
[waggle.lic] v0.18.4

mrhoribu:
Update wander proc for FalseClass error


LeBigNasty:
Really should roll this into ;ewaggle

mrhoribu:
> Really should roll this into ;ewaggle

I do agree.

mrhoribu:
> Really should roll this into ;ewaggle

We can roll this in and then I can merge this into the ewaggle PR cod #675

PR #1137, opened 2023-04-07
[gameobj] fix boons

mrhoribu:
Update boon creatures to only apply to actual boon creatures


mrhoribu:
Updated list to match up with https://gswiki.play.net/Category:Boss_creatures creatures minus newer areas as they were done properly via their own migration files.

PR #1148, opened 2023-04-26
Update eloot.lic

ServerFailure:
Allows looting of cursed if wearing an eonake gauntlet.  https://gswiki.play.net/Eonake_gauntlet


LeBigNasty:
Resolved in #1152

PR #1156, opened 2023-05-17
[effect-list.xml] initial creation for ownership

mrhoribu:
Create new XML for conversion of spell-list.xml to gain ;repo ownership.


OSXLich-Doug:
I believe line 57 in `jinx.lic` might need a look:

```    def self.out(msg, label: :debug)
      return _write _view(msg, label) unless msg.is_a?(Exception)
      ## pretty-print exception```

Pretty sure that should be `unless msg.is_a?(StandardError)` given the modification you made.

PR #1157, opened 2023-05-20
[treim.lic] v2.24.2 GROUP output update

mrhoribu:
Fix for GROUP output update and Rubocop cleanup


mrhoribu:
Still needs a fix for when not leading a group to be populated correctly.

PR #1159, opened 2023-05-20
[manaleech.lic] v2.7 GROUP output fix

mrhoribu:
Fix for new GROUP output display


mrhoribu:
Still needs a fix for when not leading a group to be populated correctly.

PR #1196, opened 2023-06-25
[eforgery.lic] Fix a couple bugs, and add ClimateWear/pocket messaging support.

trisscar1212:
Noted a couple of these bugs when using eforgery and fixed them, namely, a catch for when the script should manually go to the door or grinder, and when a glyph was missing.

In addition, I have a ClimateWear cloak, and the messaging for interacting with this container is a little different, so I added those success keywords to the you_get and you_put methods.

### DIFF
--- a/scripts/eforgery.lic
+++ b/scripts/eforgery.lic
@@ -979,7 +979,8 @@ class << forger
         done = true
         waitrt?
         tongs
-      elsif line =~ /Roundtime: 30 sec./ # didn't glance at the grinder or the door; time to do this manually
+      elsif line =~ /Roundtime:/ # didn't glance at the grinder or the door; time to do this manually.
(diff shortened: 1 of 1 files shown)


scripts/eforgery.lic
```diff
@@ -799,16 +799,16 @@ class << forger
...
       end
       fput "close #{container}" if closed
     else
       echo "container doesn't exist" if @DEBUG
       fput "get my #{item}"
-      line = waitfor "You remove", "It's closed", "Get what"
+      line = waitfor "You remove", "You grab", "It's closed", "Get what"
     end
-    return line =~ /You remove/
+    return line =~ /You remove/ ||  line =~ /You grab/
```
trisscar1212, line 811: Ah, extra space after the `||`, will fix

mrhoribu (approved):

PR #1205, opened 2023-06-30
Add localchat.lic - gtk3 window for isolating chat messages

strnglp:
Useful for clients other than Wrayth, such as Avalon.


strnglp:
We did it, we defeated Robocop!

PR #1211, opened 2023-07-05
[localchat.lic] refactor and minor bugfixes, details in header

strnglp:
- chore: refactor loop for simplicity and to be more rubyish
- bugfix: removed unintentional delay in closing the window with the close button
- bugfix: no longer displays non-dialog lines through better synchronization

### DIFF
--- a/scripts/localchat.lic
+++ b/scripts/localchat.lic
@@ -6,12 +6,14 @@
           game: Gemstone
           tags: chat
       required: Lich >= 5.0
-       version: 1.0.2
+       version: 1.0.3
 
      Help Contribute: https://github.com/elanthia-online/scripts
 
   Version Control:
     Major_change.feature_addition.bugfix
+  v1.0.3 (2023-07-06)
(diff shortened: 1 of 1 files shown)


scripts/localchat.lic
```diff
@@ -16,9 +16,15 @@
     - bugfix for default dialog text
   v1.0.0 (2023-07-01)
     - initial release
+  v1.0.1 (2023-07-04)
+    - chore: refactor loop for simplicity and to be more rubyish
+    - bugfix: removed unintentional delay in closing the window with the close button
+    - bugfix: no longer displays non-dialog lines through better synchronization
```
mrhoribu, line 22: This should really be 1.0.2 and update the version number above it. Otherwise, all looks good!

mrhoribu (changes requested):

scripts/localchat.lic
strnglp: Right you are, thank you!

mrhoribu (approved):

PR #1222, opened 2023-07-14
[ebounty] v1.1.25 misc bugfixes, heirloom and HW guard

Deysh:


scripts/ebounty.lic
```diff
@@ -178,9 +181,9 @@
 =end
 
 require 'yaml'
+SCRIPT_NAME = script.name
```
mrhoribu, line 184: This should likely be `Script.current.name` to be namescoped properly with the `Script` api.

Deysh: I'll remove that. I forgot that Script.current.name would return the script name anywhere in the script

PR #1227, opened 2023-07-19
[bigshot.lic] v5.0.0 updates to method names and code reordering

Deysh:


mrhoribu:
This is a fairly substantial update to bigshot with changes to the various method names due to case changing. This really should be a MAJOR version bump, not just a bugfix or feature addition number change for the version. Just throwing that out there. It should really be v5.0.0

Deysh:
fine by me

PR #1230, opened 2023-07-20
[bigshot.lic] handle roundtime for rogue guild cmans

FarFigNewGut:
gotta go fast....but not that fast


Deysh:
incorporated into #1227

PR #1256, opened 2023-08-25
[eloot]v1.6.3 update for dropping boxes in shattered, excluded shimme…

Deysh:
…ring orbs from sell list, misc bug fixes

### DIFF
--- a/scripts/eloot.lic
+++ b/scripts/eloot.lic
@@ -4170,7 +4170,7 @@ module ELoot
         next if $sell_ignore.include?(thing.id)
         next if ELoot.data.ready_list.find { |_k, v| v.id == thing.id }
         next if !ELoot.data.settings[:sell_exclude].empty? && thing.name =~ /#{ELoot.data.settings[:sell_exclude].join('|')}/
-        next if thing.name =~ /bound|shimmering \w+ orb/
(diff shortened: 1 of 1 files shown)


scripts/eloot.lic
```diff
@@ -4160,7 +4170,7 @@ module ELoot
         next if $sell_ignore.include?(thing.id)
         next if ELoot.data.ready_list.find { |_k, v| v.id == thing.id }
         next if !ELoot.data.settings[:sell_exclude].empty? && thing.name =~ /#{ELoot.data.settings[:sell_exclude].join('|')}/
-        next if thing.name =~ /bound/
+        next if thing.name =~ /bound|shimmering \w+ orb/
```
mrhoribu, line 4173: Wondering if these should be wrapped in appropriate anchors, so becomes:
```ruby
next if thing.name =~ /\bbound\b|^shimmering \w+ orb$/

mrhoribu (changes requested):
Just a minor thing, but otherwise, good changes!

scripts/eloot.lic
Deysh: Agree, that makes sense.

mrhoribu (approved):
lgtm

PR #1260, opened 2023-08-26
[bigshot.lic] use Spell.force_incant instead of bs_put

FarFigNewGut:


mrhoribu:
Already addressed in PR #1258

PR #1263, opened 2023-09-01
Debt collector fixes

emcguinness:
Previously if in the bank at the time another players debt was collected it was being tracked as withdrawn form your own account. Also added fix for when you do not have enough to cover the whole fine, without the full fine was being deducted.

### DIFF
--- a/scripts/ledger.lic
+++ b/scripts/ledger.lic
@@ -11,7 +11,7 @@
    contributors: Ondreian, Tysong
            game: Gemstone
            tags: silver, bounty, ledger, bank
-        version: 1.3.0
+        version: 1.3.1
    requirements:
     - sequel gem
     - ascii_charts gem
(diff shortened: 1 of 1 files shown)


emcguinness:
Not the cleanest of changes, this was a bit of a least changes approach. I did start working on changing regex's to use named capture groups and moved to an array of regex rather than one long one to save scrolling but felt I should move that to a seperate  change and test it properly as it is a sizeable change. Would make maintaining this easier going forward though.

scripts/ledger.lic
```diff
@@ -226,9 +230,20 @@ module Ledger
...
           withdraws = $1.delete(',').to_i + $4.delete(',').to_i + $5.delete(',').to_i + $6.delete(',').to_i + $7.delete(',').to_i + $10.delete(',').to_i
           deposits = $2.delete(',').to_i + $3.delete(',').to_i + $8.delete(',').to_i + $9.delete(',').to_i + $11.delete(',').to_i
+
+          if line.include?("I have a bill of ")
+            # 5 lines is the min, using 10 for safety. Very unlikely of a false positive
+            next unless reget 10, "The local debt collector suddenly enters and walks toward you purposely."
+
+            if (reget 10, "The debt collector reminds you that you still owe")[-1] =~ /owe ([\d,]+) silver/
```
ondreian, line 241: please use `.last` here and ruby parentheses, this is not very Ruby-esque
```ruby
reget(10, "The debt collector reminds you that you still owe").last
```

ondreian (changes requested):

ondreian (approved):
lgtm

PR #1273, opened 2023-09-05
[ledger.lic] v1.3.2 code optimization & cleanup

emcguinness:
This is intended to be an entirely non-function change, ready for future functional changes, as the regex's were quite hard to work with in previous form. All the old regex's are confirmed matching correctly post change (and majority of them are unchanged anyway).

If there is a better way to do this, let me know. And just for reference reason for this was to make it easier to re-introduce note tracking (currently only works in Terras), thought it made sense to split into two smaller releases for easier review (happy to bundle it all into one though)

### DIFF
--- a/scripts/ledger.lic
+++ b/scripts/ledger.lic
@@ -54,6 +54,27 @@ unless failed_to_load.empty?
   exit
 end
 
+
+WITHDRAW_SILVER_REGEX = [
+  /I have a (?<debt>bill) of (?<silver>[\d,]+) silvers? (?:presented by your creditors|that I suggest you pay immediately)/,
+  /Very well, a withdrawal of (?<silver>[\d,]+) silver/,
+  /teller scribbles the transaction into a book and hands you (?<silver>[\d,]+) silver/,
+  /teller carefully records the transaction, (?:and then )?hands you (?<silver>[\d,]+) silver/,
+  /scrip for (?<silver>[\d,]+) silvers?, with a 0 silver fee for the scrip/, # Terras note withdraw, will be moved to notes regex next release
+
+]
+WITHDRAW_NOTES_REGEX = [ # Unused till next release
+  /scrip for (?<silver>[\d,]+) silvers?, with a (?<fee>[\d,]+) silvers? fee for the scrip/, # Terras note withdraw
+]
+
+DEPOSIT_REGEX = [
+  /You deposit (?<silver>[\d,]+) silvers? into your account/,
+  /That's a total of (?<silver>[\d,]+) silver/,
+  /That's (?<silver>[\d,]+) silver|silvers? to your account/,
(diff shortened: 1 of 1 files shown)


scripts/ledger.lic
```diff
@@ -54,6 +56,27 @@ unless failed_to_load.empty?
   exit
 end
 
+
+WITHDRAW_SILVER_REGEX = [
+  /I have a (?<debt>bill) of (?<silver>[\d,]+) silvers? (?:presented by your creditors|that I suggest you pay immediately)/,
+  /Very well, a withdrawal of (?<silver>[\d,]+) silver/,
+  /teller scribbles the transaction into a book and hands you (?<silver>[\d,]+) silver/,
+  /teller carefully records the transaction, (?:and then )?hands you (?<silver>[\d,]+) silver/,
+  /scrip for (?<silver>[\d,]+) silvers?, with a 0 silver fee for the scrip/, # Terras note withdraw, will be moved to notes regex next release
```
emcguinness, line 65: This regex is the most changes, used to match x,xxx silver fee but as all fees are 0 now this simplifies things. Also does not really fit here as this regex is for silvers, will be replaced by below WITHDRAW_NOTES_REGEX but left here to prevent regressions before that.

mrhoribu: These should exist inside the module, not outside it. Could use @withdraw_notes_regex as the name as an example and reference it accordingly. That keeps things scoped into the Ledger namespace.

mrhoribu (changes requested):

scripts/ledger.lic
mrhoribu: Also you can prevent having to use Regexp.union in the matching lines and instead define the variables with it instead. so it becomes:
```ruby
@withdraw_silver_regex = Regexp.union(
  /I have a (?<debt>bill) of (?<silver>[\d,]+) silvers? (?:presented by your creditors|that I suggest you pay immediately)/,
  /Very well, a withdrawal of (?<silver>[\d,]+) silver/,
  /teller scribbles the transaction into a book and hands you (?<silver>[\d,]+) silver/,
  /teller carefully records the transaction, (?:and then )?hands you (?<silver>[\d,]+) silver/,
  /scrip for (?<silver>[\d,]+) silvers?, with a 0 silver fee for the scrip/, # Terras note withdraw, will be moved to notes regex next release
)
```

mrhoribu:
I also updated the PR name from `Cleanup` to `[ledger.lic] v1.3.2 code optimization & cleanup` to be easier to understand what the PR is for and match other PR naming standards for the repository.

emcguinness:
Thanks, definitely makes sense. Changes made.

mrhoribu (approved):
been testing and working fine. LGTM

PR #1276, opened 2023-09-07
[ledger.lic] v1.3.4 note tracking

emcguinness:
This feature adds tracking of withdrawing notes from the bank, previously this only worked on Terras where gameline includes value of note withdraw:
`The teller nods and says, "Let's see... that's a scrip for 101 silvers, with a 0 silver fee for the scrip. `

In other towns the value is not shown:
`The teller purses her lips and says, "Hmm, we will have to give you a City-States promissory note for that amount."  The teller makes some marks on a blank note and hands it to you, saying "That is a 0 silver surcharge.  Have a nice day."`

This change enables the XML feed to get the ID of note being passed to you and uses `Lich::Util::quiet_command_xml` to read the note and extract the value. Known issue with this approach is if 2 or more notes are withdrawn before the first note can be read will cause notes to be missed.

Possible alternatives to fix known issue:

1. Don't use a quite_command, just read the note 'out loud' (keep track of it's ID) and add an extra if check to the main while loop to read record the result of the read (if note ID matches)
2. Something with down stream hooks might work for keeping the read quiet, not really thought through fully as quickly got messy

Option 1 seems easy but having the read visible might annoy/confuse people? Sure there is a better way though that I am missing which is reason for mentioning it.

### DIFF
--- a/scripts/ledger.lic
+++ b/scripts/ledger.lic
@@ -39,7 +39,9 @@
       - fix to match debts in mist harbor.
     v1.3.2 - 2023-09-06
       - Code cleanup, no functional changes
-    v1.3.3 - 2023-09-08
+    v1.3.3 - 2023-09-17
+      - Fix to bounty point tracking (broken in 1.3.2)
+    v1.3.4 - 2023-09-18
       - Add tracking of note withdrawal (Withdrawing 2 or more notes rapidly before first note is tracked will cause notes to be missed)
 =end
 
@@ -314,7 +316,7 @@ module Ledger
           silver += fee
           echo "recorded.withdraw : #{silver}"
           self.record_transaction(amount: -silver, type: "silver")
-        elsif line =~ /\[You have earned (<?amount>[\d,]+) bounty points/
+        elsif /\[You have earned (?<amount>[\d,]+) bounty points/ =~ line
           self.record_transaction(amount: amount.delete(',').to_i, type: "bounty")
         elsif line =~ /inter-town bank transfer options? available/
           self.print()
--- a/scripts/ledger.lic
+++ b/scripts/ledger.lic
@@ -105,20 +105,21 @@ module Ledger
   end
 
   module Utility
-    def self.reget_xml(*lines) # Match original reget
-      status_tags
-      matches = reget lines
-      status_tags
-      return matches
-    end
-
     def self.get_note_value(noteId)
-      # line = dothistimeout "read ##{noteId}", 3, /and reads, "Hold in right hand to use."/
       line = Lich::Util::quiet_command_xml("read ##{noteId}", /has a value of (?:[\d,]+) silver and reads/, /<prompt/, false, 3).last
-      echo line
       if /has a value of (?<silver>[\d,]+) silver/ =~ line
         return silver.delete(',').to_i
       end
+    end
+
+    def self.strip_xml(line) # Copied from reget function
+      line.gsub!(/<pushStream id=["'](?:spellfront|inv|bounty|society)["'][^>]*\/>.*?<popStream[^>]*>/m, '')
+      line.gsub!(/<stream id="Spells">.*?<\/stream>/m, '')
+      line.gsub!(/<(compDef|inv|component|right|left|spell|prompt)[^>]*>.*?<\/\1>/m, '')
+      line.gsub!(/<[^>]+>/, '')
+      line.gsub!('&gt;', '>')
+      line.gsub!('&lt;', '<')
+      return line
     end
   end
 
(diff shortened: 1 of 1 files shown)


scripts/ledger.lic
```diff
@@ -98,6 +106,25 @@ module Ledger
     end
   end
 
+  module Utility
```
emcguinness, line 109: Not sure if these functions were really worth pulling into own module, was more when I added but got removed after I made further changes

scripts/ledger.lic
```diff
@@ -250,7 +277,9 @@ module Ledger
     end
 
     def self.main()
-      while (line = get)
+      status_tags
+      while (xml_line = get)
+        line = Ledger::Utility.strip_xml(xml_line.dup)
```
emcguinness, line 282: Most regex's worked fine without xml, only needed for extracting ID of note handed to you so rather than change all regex to XML version felt easier to make xml and non xml game lines available for match.

scripts/ledger.lic
```diff
@@ -269,6 +298,22 @@ module Ledger
...
           self.record_transaction(amount: silver, type: "silver")
+        elsif (match = line.match(@withdraw_notes_regex))
+          silver = match[:silver].delete(',').to_i
+          fee = match[:fee].delete(',').to_i
+          silver += fee
+          echo "recorded.withdraw : #{silver}"
+          self.record_transaction(amount: -silver, type: "silver")
+        elsif (match = xml_line.match(@withdraw_notes_xml_regex))
+          fee = match[:fee].delete(',').to_i
+          silver = Ledger::Utility.get_note_value(match[:noteId]) # If further notes are withdrawn before this returns they will be missed
```
emcguinness, line 309: Mentioned in opening comment but during this note read any further note withdraws will be missed.

scripts/ledger.lic
```diff
@@ -98,6 +106,25 @@ module Ledger
     end
   end
 
+  module Utility
+    def self.get_note_value(noteId)
+      line = Lich::Util::quiet_command_xml("read ##{noteId}", /has a value of (?:[\d,]+) silver and reads/, /<prompt/, false, 3).last
```
emcguinness, line 111: Thinking now this was a 5.7.0 thing so would delay release until out of beta (and possibly a while after to allow time to update). If this is a 5.7 thing is there a good alternative? Could pull the issue_command code directly into the script I guess.

scripts/ledger.lic
mrhoribu: Should be fine to leave as a sub-module. It makes sense to me and more organization the better in my book.

scripts/ledger.lic
```diff
@@ -98,6 +106,25 @@ module Ledger
     end
   end
 
+  module Utility
+    def self.get_note_value(noteId)
```
mrhoribu, line 110: use snake_case_for_variables instead of CapilizationCasingOfVariables. That's the preferred ruby best practice.

mrhoribu:
Have you thought about instead of reading the note to check the upstream buffer to see what the most recent WITHDRAW command was? Granted, that could get complex considering I'm not sure if people do `w 100000` or what have you at the bank or what method to try to match against.

scripts/ledger.lic
mrhoribu: Yep, quiet_command_xml is the way to go for now until 5.7.0 is released. Can always change it later on once 5.7.0 has had time to bake in for awhile. Can also add a requires in the header and code at the top to warn if not on v5.7.0+ and exit out.

scripts/ledger.lic
mrhoribu: Uncertain how I feel about using strip_xml vs just updating the regex to appropriate XML version equivalent. Ends up being the same in the end.

scripts/ledger.lic
mrhoribu: Have you thought about instead of reading the note to check the upstream buffer to see what the most recent WITHDRAW command was? Granted, that could get complex considering I'm not sure if people do w 100000 or what have you at the bank or what method to try to match against.

emcguinness:
Yeh, monitoring upstream was one of my first thoughts I just could not think of a good way to correlate a command to an actual in game action. Just cos they typed a valid withdraw command there is no guarantee it succeeded, would be struck trying to correlate just based on timing which could cause more issues (at least that was my thought).

scripts/ledger.lic
emcguinness: Damn, thought I had finally beat the habit...

scripts/ledger.lic
emcguinness: This way was more me being lazy, at this point it would not take much just to grab the xml versions for the rest. They could likely do with cleanup anyway as they were a bit piecemeal as the list grew over time no doubt and I never validated all the existing ones in game just copied them to a new home and tested via ;send to make sure everything would trigger.

emcguinness:
Mostly thinking out loud here but we can't check gameline for withdraw and then watch the upstream as already missed it. Options:

1. Change while=get loop to get? so it is non blocking, create small FIFO queue of any commands that look like a withdraw (aliases would be missed) along with a timestamp and then when you see a withdraw hope the "oldest but still recent" one in the queue is valid? 
2. Upstream hook, guess this would be much the same as the first just now we have two threads to play with so no need to change the blocking get.
3. ??

While writing this I remembered the thing I forgot, what if a script is withdrawing? From my quick test that would not be seen in upstream_get, it is not common I think but I know I do occasionally with the ;gbank script.

mrhoribu:
Been running this for a few days, no issue so far. So going go to ahead and merge it in.

PR #1282, opened 2023-09-12
[repo] V2.53 - Add wrayth links to repo output 

emcguinness:
Initial draft to show a possible implementation. Before this change a ;repo download action pauses for 3 seconds before continuing, due to ease of accidentally clicking a link logic is added to prompt for user confirmation via ;send and if not given in time the script will exit without downloading.

This is achieved by adding a '--clicked' link to all links the script generates, this means any manual ;repo download commands will still exhibit old behaviour. This means new feature (links) new behaviour (;send confirmation instead of auto download after timeout). This behavior could become the default for all "dangerous" actions.

In this current from the new behavior is only implemented into download action but will be added to delete and (probably) rate as that will cover the three "dangerous" actions that could be done via the links.

First download  command example of no duplicate names, second download is with duplicates (bonus text appears to explain how to download).
First appearance of ">alias.lic" is a link click, this one no response is given and times out
Second appearance of ">alias.lic" is a link click, followed by a confirmation (and download completing)
image


scripts/repository.lic
```diff
@@ -1317,10 +1341,46 @@ elsif (cmd[0] =~ /^download$/i)
...
+    if link_clicked
+      time_out = 10
+      start_time = Time.now
+      confirmed = false
+
+      respond ""
+      respond " You have clicked a link to download #{list[0][fi]} by the author #{list[0][headers.index('author')]}"
+      respond " To confirm this action please ';send yes' in the next #{time_out} seconds to confirm."
+      respond ""
+
```
emcguinness, line 1364: Below loop to be extracted as a get_user_confimration proc for reuse in other functions that need it (like delete/rate)

emcguinness:
Moved confirmation logic to proc for re-use and added to rate and delete (and addition to the download) which I think covers all the "dangerous" actions that could be performed via link clicks. Getting pretty close now but need to run though all commands to check for accidental side effects.

image

emcguinness:
Tested commands that call format_list and are affected by this change:

1. List - click for info
2. list-updates - click for info
    - Could be a case to be argued for click to download given context but info feels better to me (check info before downloading update)
3. list-new - click for info
4. info - If multiple match produces list with click for info
    - TODO: add text prompt with help for cli flags for non clicky clients
5. download -  If multiple match produces list with click for download, prompts for confirmation first
6. set-updatable - (Just added) If multiple match produces list with click to set updatable, prompts for confirmation first
    - TODO: add text prompt with help for cli flags for non clicky clients
7. delete - If multiple match produces list with click to delete, prompts for confirmation first
    - TODO: add text prompt with help for cli flags for non clicky clients
 8. rate - If multiple match produces list with click to rate, prompts for confirmation first
    - TODO: add text prompt with help for cli flags for non clicky clients

Other things of note:

1. show-updatable - As the table is done differently not affected by this change, quick POC shows it could work but throws column size off if done before table creation. Might revisit this.
2. unset-updatable - Posible slight bug (unrelated to these changes), if two or more scripts partially match the name you are trying to unset it will just remove one at a time without prompt.

> ;repo unset-updatable flet
> --- Lich: repository active.
> [repository: the download-updates command will ignore updates for fletchmaster.lic]
> --- Lich: repository has exited.
> ;repo unset-updatable flet
> --- Lich: repository active.
> [repository: the download-updates command will ignore updates for fletching.lic]

emcguinness:
Moved the "duplicate file" help text to a proc and called anywhere it is required. Moving out of draft now as think this is about there.

Final maybe todo to check on is adding links to show-updatable for completeness.

emcguinness:
Looking like a no on the show-updatable for this release. If you add the XML before creating the table it will set the column widths all wrong. There is currently no way to set the width of a single column, all columns have to be same size. There is a issue raised to fix this, maybe return once that is done.

https://github.com/tj/terminal-table/issues/64

mrhoribu:
> Looking like a no on the show-updatable for this release. If you add the XML before creating the table it will set the column widths all wrong. There is currently no way to set the width of a single column, all columns have to be same size. There is a issue raised to fix this, maybe return once that is done.
> 
> [tj/terminal-table#64](https://github.com/tj/terminal-table/issues/64)

Yep, you have to add the XML code AFTER the fact, otherwise the table spacing adjusts for the extra xml even though it's not shown on the client. You can see what I did here for `resource.lic` to allow for monsterbolding the header text
```ruby
    table = table.to_s
    if table =~ /\n\|([\w\d\s\|\$\/\(\)\.]+)\|\n/
      header = $1
      headerbold = Lich::Messaging.monsterbold(header)
      table = table.gsub(header, headerbold)
    end
    _respond "<output class=\"mono\"/>\n" + table + "\n<output class=\"\"/>"
  end
```

emcguinness:
Ok all commands I could think of tested and the last push fixes things for when using the --hide-author flag.

If there is no author but there is a name clash you will get the game field, this gets added to the link. If there is no author and no name clash then we have nothing but the name left, this will still get turned into a link. I think in this final case there is likely a chance of a clash but as the clicky links do not have the --hide-author flag the problem will correct itself with worse case scenario of one extra click.

This point I think I am done with anything I can think of, so just any final comments/suggestions.

mrhoribu:
This should be good now for full testing. Updated output listings to be in mono format.

PR #1299, opened 2023-09-21
[TargetWindow.lic] v1.3 - Arm Count

emcguinness:
Minor addition intended for sorcerers to show the total amount of arms in the room.
Window:
image

Quick:
image
Honestly not sure if I like how it looks in the quick bar, but I guess it is also optional.

Also as mentioned if you don't want these changes no pressure to accept them, as the Sorcerer I liked the removal of the spam of arms but still like to know how many I have left.

### DIFF
--- a/scripts/targetwindow.lic
+++ b/scripts/targetwindow.lic
@@ -47,14 +47,21 @@ module TargetWindow
     old_targets = GameObj.targets
     _respond("<closeDialog id='TargetWindow'/><openDialog type='dynamic' id='TargetWindow' title='Targets' target='TargetWindow' scroll='manual' location='main' justify='3' height='68' resident='true'><dialogData id='Targets'></dialogData></openDialog>")
     loop {
-      if GameObj.targets.length > 0
+      targets = GameObj.targets.select { |t| !(t.noun =~ @grasp_arms || t.name =~ /^animated /) }
+      if targets.length > 0
         output = "<dialogData id='TargetWindow' clear='t'>"
-        output += "<link id='total' value='Total Targets: #{GameObj.targets.delete_if { |t| t.noun =~ @grasp_arms }.length}' cmd='target next' echo='target next' justify='3' top='3' left='0' height='15' width='195'/>"
-        GameObj.targets.each { |t|
-          next if t.noun =~ @grasp_arms
+        output += "<link id='total' value='Total Targets: #{targets.length}' cmd='target next' echo='target next' justify='3' top='3' left='0' height='15' width='195'/>"
+        targets.each { |t|
           target_status = TargetWindow.status_fix(t.status) unless t.status.nil?
           output += "<link id='#{t.id}' value='#{target_status} #{t.name.split.map(&:capitalize).join(' ')}' cmd='target ##{t.id}' echo='target ##{t.id}'  justify='3' left='0' height='15' width='195'/>"
         }
+        echo "setting #{CharSettings['show_arms']}"
+        if CharSettings['show_arms']
+          echo "arms!!"
+          arms = GameObj.targets.select { |t| t.noun =~ @grasp_arms }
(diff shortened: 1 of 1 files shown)


scripts/targetwindow.lic
```diff
@@ -104,9 +115,16 @@ elsif Script.current.vars[0].to_s =~ /^flash/
     _respond("Current target will not flash.")
     exit
   end
+elsif Script.current.vars[0].to_s =~ /^showarms/
```
emcguinness, line 118: Not sure I like the name of the flag, open to suggestions

scripts/targetwindow.lic
```diff
@@ -47,14 +49,19 @@ module TargetWindow
     old_targets = GameObj.targets
     _respond("<closeDialog id='TargetWindow'/><openDialog type='dynamic' id='TargetWindow' title='Targets' target='TargetWindow' scroll='manual' location='main' justify='3' height='68' resident='true'><dialogData id='Targets'></dialogData></openDialog>")
     loop {
-      if GameObj.targets.length > 0
+      targets = GameObj.targets.select { |t| !(t.noun =~ @grasp_arms || t.name =~ /^animated /) }
```
emcguinness, line 52: Moved the logic to strip unwanted things to the top as it was getting a bit messy with needing to check for more things further down with the addition of the animate check. I also have not tested a bard to know if this matches (think it does) but doubt they would want to see their weapon here anyway.

scripts/targetwindow.lic
```diff
@@ -47,14 +49,19 @@ module TargetWindow
...
-        GameObj.targets.each { |t|
-          next if t.noun =~ @grasp_arms
+        output += "<link id='total' value='Total Targets: #{targets.length}' cmd='target next' echo='target next' justify='3' top='3' left='0' height='15' width='195'/>"
+        targets.each { |t|
           target_status = TargetWindow.status_fix(t.status) unless t.status.nil?
           output += "<link id='#{t.id}' value='#{target_status} #{t.name.split.map(&:capitalize).join(' ')}' cmd='target ##{t.id}' echo='target ##{t.id}'  justify='3' left='0' height='15' width='195'/>"
         }
+        if CharSettings['show_arms']
+          arms = GameObj.targets.select { |t| t.noun =~ @grasp_arms }
+          output += "<label id= 'armCount' value='Total arms: #{arms.length}' top='#{15 * (targets.length + 2)}' left='0'  width='187'/>" unless arms.empty?
```
emcguinness, line 62: The "top" section is basically dynamically adding 2 lines of space for the "Total arms:" after the mob list.

scripts/targetwindow.lic
emcguinness: Oh forgot the mean reason for this was without we would get "targets: 0" if there were only excluded targets in the room rather than then the "no targets" text

scripts/targetwindow.lic
```diff
@@ -68,14 +75,18 @@ module TargetWindow
...
         output = "<dialogData id='quick' clear='t'>"
-        output += "<link id='twtotal' value='Total Targets: #{GameObj.targets.delete_if { |t| t.noun =~ @grasp_arms }.length}' cmd='target next' echo='target next'/>"
-        GameObj.targets.each { |t|
-          next if t.noun =~ @grasp_arms
+        output += "<link id='twtotal' value='Total Targets: #{targets.length}' cmd='target next' echo='target next'/>"
+        targets.each { |t|
           target_status = TargetWindow.status_fix(t.status) unless t.status.nil?
           output += "<link id='#{t.id}' value='#{target_status} #{t.name.split.map(&:capitalize).join(' ')}  ' cmd='target ##{t.id}' echo='target ##{t.id}'/>"
         }
+        if CharSettings['show_arms']
```
emcguinness, line 86: Is checking directly against CharSettings inside a loop bad practice? This could be assigned to a var before the loop starts. Not sure if this is causing extra DB lookups each loop when this value is not likely to change (the only way it could be changed while script is running is via ;force.

PR #1300, opened 2023-09-21
[Multiple] - Bugfix - Change in Wealth output

emcguinness:
The output of the wealth command changed today:
https://discord.com/channels/226045346399256576/1121832778791665716/1154365026942070835

I have done a quick search for all EO scripts implementing their own wealth check and update the regex, is a chance I missed some but this should be a good start. I have not been able to test all of the script myself but have tried eherb/eloot and they seem ok and the rest implement in basically the same way.

### DIFF
--- a/scripts/coins.lic
+++ b/scripts/coins.lic
@@ -27,7 +27,7 @@
 person = variable[1]
 keep_coins = variable[2].to_i
 
-loop { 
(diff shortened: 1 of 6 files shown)


mrhoribu (approved):
lgtm

PR #1330, opened 2023-10-09
[OverWatch.lic] silent strike update

snperry85:
Added a check to see if target re-hid on events that could be a silent strike. Not adding to GameObj if a hide was successfully detected after an attack.

### DIFF
--- a/scripts/overwatch.lic
+++ b/scripts/overwatch.lic
@@ -64,7 +64,7 @@ module OverWatch
       end
     end
     echo "Silent Strike not detected or failed." if silent_strike && @debug
-    
+
     unless GameObj.targets.any? { |npc| npc.id == target_id }
       GameObj.new_npc(target_id, target_noun, target_name.gsub(/  /, " "))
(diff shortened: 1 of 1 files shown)


scripts/overwatch.lic
```diff
@@ -41,21 +44,34 @@ module OverWatch
...
+    unless GameObj.targets.any? { |npc| npc.id == target_id }
       GameObj.new_npc(target_id, target_noun, target_name.gsub(/  /, " "))
+      echo "#{target_id} added to GameObj.npcs."
     end
-    if XMLData.current_target_ids.include?(target_id)
-      echo "ID already in current_target_ids" if @debug
-    else
+    unless XMLData.current_target_ids.include?(target_id)
       XMLData.current_target_ids.unshift(target_id)
+      echo "#{target_id} added to XMLData.current_target_ids."
```
mrhoribu, line 74: Did you mean to leave in the echos without @debug being set?

snperry85: I did not. Thanks

PR #1345, opened 2023-11-02
[eloot.lic] check mapdb and XMLData room desc for trashcan fallback

FarFigNewGut:


FarFigNewGut:
just for context it looks like the IMT trashcan was changed from "old beer table" to "old oak beer table" (the mapdb should be updated now for the meta tag)

But this fallback logic if the tag fails didn't work either because it was checking the Room.current.description (mapdb) and the mapdb wasn't updated to add the table to the desc so it didn't see it. I'm assuming it was checking the mapdb and not GameObj there because of maybe race conditions loading the rooms with a lot of objects we saw? or just an oversight, not sure. Either way there's no harm in checking both since it's doing a scan + uniq so even if it finds the object twice it's fine. Will make it more robust to other potentially  changed/new rooms.

mrhoribu (approved):

PR #1360, opened 2023-11-27
Update kswole.lic

HebrewToYou:
- Adding support for some IMT spell preps
- Checking to ensure that FEAT ABSORB is not on cooldown
- Checking for 30 stamina or more before using FEAT DISPEL


mrhoribu (approved):
The cooldown issue is due to a bug in Lich 5.7.0's `Lich::Util.normalize_lookup` method that should get patched in 5.7.1 release. The fix in this script should be good to go till then. The DISPEL MAGIC should already be checking for 30stam but an additional check doesn't hurt it. Otherwise, this looks good!

PR #1361, opened 2023-11-29
[bigshot] v5.3.0 added wander guard and 2nd Rift fog casting

Deysh:


snperry85:
For some reason `$bigshot_briar` is never getting set to true, even when changing the regex to match what is in effects-list.xml. I can't figure out the reason. This causes it to trigger the buff via `raise` and use all your influence as soon as it reaches 100, removing chances for briar flares to trigger. Since the buff has been added to effects-list and the previous spell-list I don't see a reason for bigshot's stand alone tracking anymore.

https://github.com/elanthia-online/scripts/blob/016906e7db4fddb11c1115e81486982db192afef/scripts/bigshot.lic#L3306
We can swap this `$bigshot_briar` to `Spell[9105].active?`

I loaded up lich 5.6.2 with infomon 1.18.11 and the tracking works properly. So I don't think there will be any backwards compatibility issues.

This will also allow us to remove the following portions to dealing with `$bigshot_briar`
https://github.com/elanthia-online/scripts/blob/d88156bbd48e76c4a7de632877f70ed93e3e0b9f/scripts/bigshot.lic#L230
https://github.com/elanthia-online/scripts/blob/d88156bbd48e76c4a7de632877f70ed93e3e0b9f/scripts/bigshot.lic#L1692C1-L1695C32

Deysh:
Thanks for the catch. Changes submitted.

Issue #1368, opened 2023-12-07
[script] ebounty not matching herbs correctly


bmg90001:
### Script Name

ebounty - wrong forage name wild phlox

### Bug Report

ebounty forages for sprig of wild phlox, and the game is expecting wild phlox to be foraged. This might apply to all of the sprig of herbs.

### Expected Behavior

Shouldn't use sprig of, only wild phlox

### Current Behavior

searches for sprig of wild phlox which returns no herbs

### Possible Solution

?

Deysh:
the look-up table for herbs isn't used. I left it there in case something like this came up.

For the last several versions ebounty uses the XML to determine the correct command to forage.
> forage_link = Regexp.last_match(1) if lines.any? { |l| l =~ /concoction that requires <d cmd=\"(.*?)\">/ }

It could be my regex match is off or the game is sending something strange. Are you able to capture the XML if you type bounty in game while having one of those tasks? You can use ;logxml

bmg90001:
The herbalist in Ta'Vaalor, Maraene, is working on a concoction that requires <d cmd="forage sprig of jasmine">a sprig of jasmine</d> found in the Glaise Cnoc Cemetery near Ta'Vaalor.  These samples must be in pristine condition.  You have been tasked to retrieve 8 samples.
<prompt time="1701955171">&gt;</prompt>
<dialogData id='expr'><label id='yourLvl' value='Level 6' top='0' left='0' align='n' width='160' height='15'/><progressBar id='mindState' value='50' text='clear' top='45' left='3' align='n' width='160' height='15'/><progressBar id='nextLvlPB' value='53' text='7121 until next level' top='20' left='3' align='n' width='160' height='15'/></dialogData><prompt time="1701955201">&gt;</prompt>
<component id='room players'>Also here: <a exist="-11144768" noun="Eckius">Eckius</a></component>
<a exist="-11144768" noun="Eckius">Eckius</a> just arrived.
<prompt time="1701955220">&gt;</prompt>

bmg90001:
[2023-12-07-07-18-05.txt](https://github.com/elanthia-online/scripts/files/13600654/2023-12-07-07-18-05.txt)

Deysh:
Thanks. I'll look into it.

Deysh:
not able to replicate the issue....

kneel
You kneel down.
K>forage sprig of jasmine
d100(Open): 7
You forage briefly and manage to find a sprig of jasmine!
[...]

PR #1369, opened 2023-12-07
[ebounty.lic] remove sprig of from forage names

bmg90001:
remove sprig of from forage names


Deysh:
Please see: https://github.com/elanthia-online/scripts/issues/1368#issuecomment-1845167345

Deysh:
That section was legacy code.

PR #1386, opened 2023-12-18
[eloot] v1.6.15 added breakable looting and selling options

Deysh:


scripts/eloot.lic
```diff
@@ -1368,10 +1373,10 @@ module ELoot
 
   def self.read_note
     result = 0
-    note = [GameObj.right_hand, GameObj.left_hand].find { |i| i.noun =~ /note|scrip|chit/ }
+    note = [GameObj.right_hand, GameObj.left_hand].find { |i| i.noun =~ /(note|scrip|chit)\b/ }
```
mrhoribu, line 1376: Probably better to wrap it on both sides with `\b` and make the parenthesis group non-capture. Or, can change it to just be a string compare?
```ruby
    note = [GameObj.right_hand, GameObj.left_hand].find { |i| ['note', 'scrip', 'chit'].include?(i.noun) }
```

Deysh: updated it with a word boundary on each side and made it a non-capture group

PR #1415, opened 2024-01-09
[foreach.lic] correct filter regex capture groups

FarFigNewGut:
During a recent refactor the regex capture group for the main filter (m=blah, n=blah, etc.) was changed from match[2] to match[1]. This breaks the logic and makes all filters default to a TYPE filter.

Link to previous versions of script using the [2] match group (first group is either NIL or the conditional skip/first/etc. extra filters)
https://github.com/elanthia-online/scripts/blob/a836db6f0c295a0934297f02d035c3a21d2998c7/scripts/foreach.lic#L1689

### DIFF
--- a/scripts/foreach.lic
+++ b/scripts/foreach.lic
@@ -25,9 +25,11 @@
           game: Gemstone
           tags: utility, inventory,
       required: Lich > 5.0.1
-       version: 1.0.6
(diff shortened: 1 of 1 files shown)


mrhoribu (approved):
reviewed and indeed this was an error in the number used. Good catch.

PR #1430, opened 2024-02-19
[bigshot.lic] dhurl update

ServerFailure:
This updates dhurl (Hurling) to include aiming using the ambush settings.


ServerFailure:
> I'm not a hurler, but is there any difference in RT/skill checks between doing an AIM'd hurl and a non-aimed one? Incase there is, I could see issues with this routine now forcing an aimed skill check/RT addition when it didn't before.

An AIMED hurl is +1s RT.  


    if ($bigshot_ambush >= command.length())
      # If you have gotten here, there is nothing left to target on your list.
      $bigshot_ambush = 0
      command = ["chest"]
    end
    
    command = nil?

PR #1435, opened 2024-02-25
[ebounty.lic] bug fix ebounty may invoke eherbs repeatedly

gs4-Mertyn:
A young player's health may be less than 95% but still higher than the eherbs 'Hypochondriasis' condition for (checkhealth + 7) < maxhealth. This state results in ebounty repeatedly invoking the eherbs script.


gs4-Mertyn:
Updated eherbs header. Didn't change eherbs header since the only change was a code comment.

gs4-Mertyn:
Sorry, new to git, should be good now.

PR #1449, opened 2024-03-16
[eloot.lic] v1.6.20 withdraw a note when the max bank account balance is reached

gs4-Mertyn:
F2P accounts have a max bank account balance. When the cap is reached, eloot will now withdraw a note of up to 100k to make room for more silvers to be deposited into the account. It will retain at least 10k silvers in the account for running errands.


Deysh:
I added your idea into the update I was working on for the new trash verb. Appreciate the contributions.

PR #1478, opened 2024-04-25
[eloot.lic] [bigshot.lic] handle all the new custom disk nouns

FarFigNewGut:
i definitely have not had a chance to test this (will in the morning) but should handle all the million new disk nouns and make updating for new ones easier going forward. Issuing a PR in case anyone here wanted to give it a go. Also incoming rubocop violations probably.


FarFigNewGut:
i actually did get a chance to use this some last night and both bigshot and eloot worked as expected. just did some testing with the data class refactor and also working. core scripts measure twice cut thrice, etc. etc.

PR #1486, opened 2024-04-30
[map.lic] v1.1.2 File.join cleanup

mrhoribu:
update to use File.join and lich constants instead of global vars


FarFigNewGut:
LGTM

PR #1506, opened 2024-05-05
[scripts][version] Add explicit Lich Location to reported paths

MahtraDR:
Makes it a tad easier for users who need this info and might find it hard to figure it out otherwise.

### DIFF
--- a/scripts/version.lic
+++ b/scripts/version.lic
@@ -9,6 +9,8 @@
 
   Version Control:
     Major_change.feature_addition.bugfix
+  v1.0.1 (2024-05-05)
+    - add Lich Location to report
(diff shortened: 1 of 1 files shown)


MahtraDR:
image

MahtraDR:
> Isn't showing the LICH_DIR path practically the same thing as showing the LICH_DIR+lich.rbw? Feels a bit redundant to have them twice. I don't have anything against having it, just worry about `;version` getting too bloated in it's output more than is necessary causing it to get harder and harder to read thru.

There are users who can take "Look at LICH_DIR and append lich.rbw to it" and some users who will come to discord anyway, and ask us. This saves that step.

PR #1514, opened 2024-05-12
[OverWatch.lic] added kiramon stalker support

snperry85:
Added messaging for Kiramon Stalkers in the hive.
Added checks for wall of thorns, if it blocks a silent strike, it can prevent the rehide.


snperry85:
> Any reasoning to why anchoring the various regex statements was removed? Was it conflicting somehow? Usually it's better to keep things anchored if at all possible.

They were not triggering for the stalkers with the anchors. So I removed them all to prevent issues in the other areas. I don’t consider them important in this scenario as the matches are exact.

PR #1518, opened 2024-05-18
Bigshot reim time

FarFigNewGut:
Some reim mobs have a noun collision with passive NPC mobs (escorts). There was already an exception to handle reim mobs by name but it was incomplete. failing at least on the `Foreign Dignitary`

We already have all(?) the reim mobs in a migration that adds them explicitly as aggressive NPCs and adds a special `realm:reim` type to them. We should use that exception instead. I'm not aware of any non-reim mobs that would be affected by the change.

tested the new logic with:
```ruby
test_npc_nouns = ['dignitary', 'merchant', 'Dignitary']
test_npc_names = ['foreign', 'Foreign', 'half-elven']
combos = test_npc_names.product(test_npc_nouns)

objs = []
combos.each_with_index { |c,i| objs.push(GameObj.new(696969 + i, c[1], "#{c.join(' ')}")) }

objs.each { |o|
  echo o.name
}
objs.delete_if { |npc| npc.noun =~ /child|traveller|scribe|merchant|dignitary|official|magistrate/i && npc.type !~ /realm:reim/ }
echo "FILTERED:"
objs.each { |o|
  echo o.name
}
```

```
--- Lich: test active.
[test: foreign dignitary]
[test: foreign merchant]
[test: foreign Dignitary]
[test: Foreign dignitary]
[test: Foreign merchant]
[test: Foreign Dignitary]
[test: half-elven dignitary]
[test: half-elven merchant]
[test: half-elven Dignitary]
[test: FILTERED:]
[test: Foreign Dignitary]
```

We want to leave valid aggressive NPCs as the only remaining things in the list so this is the expected result (and the type check is case sensitive so `foreign dignitary` was dropped)

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -8,7 +8,7 @@
   contributors: SpiffyJr, Tillmen, Kalros, Hazado, Tysong, Athias, Falicor, Deysh, Nisugi
           game: Gemstone
           tags: hunting, bigshot, combat
-       version: 5.3.7
+       version: 5.3.8
       required: Lich >= 5.5.0, infomon >= 1.18.11
 
   Setup Instructions: https://gswiki.play.net/Script_Bigshot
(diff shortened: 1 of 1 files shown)


FarFigNewGut:
think i got all the places. i'm PRETTY sure valid_target is ok also it goes a little deep cause it's called in a few places but looks ok.

PR #1537, opened 2024-06-11
[all] [go2.lic] Move go2_start_room from sqlite UserVars to global variable

vtcifer:
Updating UserVars everytime that go2 gets called can cause some performance issues when multiboxing.

Does this value _really_ need to be persisted across sessions?  It seems to me that storing it in a global variable `$go2_start_room` in place of `UserVars.go2_start_room` would satisfy still satisfy real world use, would not suffer performance impact of UserVars.

### DIFF
--- a/scripts/go2.lic
+++ b/scripts/go2.lic
@@ -1262,15 +1262,21 @@ module Go2
   end
 
   if target_search_string =~ /goback/i
-    last_start_room = UserVars.go2_start_room.to_s
-    if last_start_room
-      target_search_string = last_start_room
-      echo "Attempting to return to last start room of: #{last_start_room}"
-    else
+    if $go2_start_room.nil?
       echo "go2 does not have a stored start room from the last time it ran."
       exit
+    else
+      last_start_room = $go2_start_room.to_s
+      if last_start_room
+        target_search_string = last_start_room
+        echo "Attempting to return to last start room of: #{last_start_room}"
+      else
+        echo "go2 does not have a stored start room from the last time it ran."
+        exit
+      end
     end
   end
+
 
   #
   # target was given; use saved settings, override them with command line settings, but don't save them
@@ -1778,7 +1784,7 @@ module Go2
   first_move   = true
 
   # Store start room for future use
-  UserVars.go2_start_room = start_room.id
+  $go2_start_room = start_room.id
 
   loop {
     moves_sent = $room_count
--- a/scripts/go2.lic
+++ b/scripts/go2.lic
@@ -1277,7 +1277,6 @@ module Go2
     end
   end
 
-
   #
   # target was given; use saved settings, override them with command line settings, but don't save them
   #


OSXLich-Doug:
In the same camp, and hopeful either an instance variable or a discrete module method can be devised.  Killing global variables is on my (lengthy!) to-do list, and for grins I scanned existing `go2`.  178 instances of `$`, 88 of which are tied to regex (which we fix by named groups) and end of regex string, which will probably stand as that is not used a variable.  Leaves about 90 primarily across 4 or so variables.  One more won't kill us, but if we're going to break with convention, we should do it locally (ala' Tysong's suggestion) instead of globally, and then having it spread when we do course correct.

vtcifer:
Just some timing data from earlier to show the delay I was referring to.

```
2024-06-11T13:53:13.966688958Z:--- Lich: go2 active.
2024-06-11T13:53:14.471511913Z:The blue-dappled prereni closes to melee range on you!
2024-06-11T13:53:18.309800703Z:--- Lich: 'textsubs' has been stopped by go2.
2024-06-11T13:53:18.310056757Z:[go2: ETA: 0:00:05 (28 rooms to move through)]
2024-06-11T13:53:18.650682740Z:[go2]>climb tree
```

5 second gap between go2 invocation and actually doing anything.

With the code change I made above:
```
2024-06-11T20:51:57.536505783Z:--- Lich: go2 active.
2024-06-11T20:51:57.536651225Z:>,go2 shard bank
2024-06-11T20:51:57.846988220Z:[go2: ETA: 0:00:03 (17 rooms to move through)]
2024-06-11T20:51:57.847164309Z:[go2]>go grounds
```

(I did see this in real time but was having an issue searching through logs at the time.  I'll go back and try to find the first go2 right after I changed it)

vtcifer:
I was thinking about this this morning and thought about all the different ways we have to store different settings, and how they're all persisted through the SQLite DB.  I think we really just should have an option to store some settings/variables/info that does not persist through sessions.  

Right now we have:
`Settings` - script scoped persistent storage that apply/are accessible regardless of character or game.  
e.g. global auto starts are stored here, as are some repository settings

`GameSettings` - script scoped persistent storage that apply/are accessible regardless character, but only when logged into specific game instance (dr vs drf vs gs4 etc). 
e.g. go2 custom targets are stored here

`CharSettings` - script scoped persistent storage that apply/are accessible only when logged into a specific character (alice vs bob) and into a specific game instance (dr vs drf vs gs4 etc). 
e.g. go2 typeaheads are stored here, lnet friends/enemies are stored here

`UserVars` - persistent storage usable by any script, but scoped to a specific character (alice vs bob) and in a specific game instance (dr vs drf vs gs4 etc).
e.g. spell auto data (how high you can cast spells), what song type you should play to learn best

What I think we should have is something like:

`SessionVars` - **non-persistent** storage usable by any script, but scoped to a specific character (alice vs bob) and in a specific game instance (dr vs drf vs gs4 etc).

We on the DR side use UserVars for things that multiple scripts need access to, but some of it really doesn't need to persist across sessions, and some does.  the spell discern data for example really needs to persist.  It CAN be recalculated, but that can be time consuming.  But some of it doesn't need to persist at all, because it either doesn't make sense to save across sessions (imo, this change for go2) as well as other things like - account type (premium, f2p, standard - we can fetch these using in game commands, and so currently at the start of every session), or moonwatch data - this is fed by moonbots for the most part from firebase.  It's only in UserVars because multiple scripts need access to it, but in fact persisting it isn't necessary because next session that data is useless unless it's updated regularly.

(I'm not sure how UserVars data is used on the GS side).

PR #1545, opened 2024-06-20
[bigshot.lic] v5.3.11 cmd_assault bugfix

mrhoribu:
fix to break from cmd_assault command on cooldown if loop repeated

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -2697,8 +2697,6 @@ class Bigshot
         break
       elsif result =~ complete_regex || Time.now() > break_out
         break
-      elsif Effects::Buffs.active?("Assaulting")
-        break
       elsif result == false
         $bigshot_should_rest = true
         $rest_reason = "Unknown result from #{cmd} routine: #{result}"


scripts/bigshot.lic
```diff
@@ -2691,8 +2693,12 @@ class Bigshot
         next
       elsif result =~ /may not be activated within 60 seconds of a Multi-Strike\./i
         break
+      elsif result =~ /(?:Barrage|Flurry|Fury|Guardant Thrusts|Pummel|Thrash) is still in cooldown\./i
```
FarFigNewGut, line 2696: If you hit a race where Effects thinks the cooldown is done but you race with the game's more precise check you won't retry so it would skip the command

scripts/bigshot.lic
```diff
@@ -2691,8 +2693,12 @@ class Bigshot
         next
       elsif result =~ /may not be activated within 60 seconds of a Multi-Strike\./i
         break
+      elsif result =~ /(?:Barrage|Flurry|Fury|Guardant Thrusts|Pummel|Thrash) is still in cooldown\./i
+        break
       elsif result =~ complete_regex || Time.now() > break_out
         break
+      elsif Effects::Buffs.active?("Assaulting")
+        break
```
FarFigNewGut, line 2701: i think you want to not break on assaulting it would be more a safety like if cooldown is active but assault is not, then break (so you know you successful did the thing at some point, but you are not actively assaulting anymore)

OSXLich-Doug:
Looks good here.

PR #1548, opened 2024-06-23
[go2.lic] v2.1.1 bugfix non-array custom targets

mrhoribu:
Bugfix to custom targets to respect non-arrays


OSXLich-Doug:
Good here

PR #1579, opened 2024-07-14
Resubmission of 1.3.1 WSL bugfix with responsiveness to resizing

strnglp:


strnglp:
Hold on this merge - I think this isn't working correctly on MacOS

strnglp:
I had to wrap the offset business in a !Mac conditional. Don't love it honestly but it seems like Mac frame_extents are somewhat inaccurate.

PR #1583, opened 2024-07-18
[map.lic] v1.3.2 WSL fixes (attempt2)

mrhoribu:
Additional fixes for WSL and responsiveness. Applied only if using WSL. See #1579 for original attempt1


strnglp (approved):
Works on WSL
Tested:
- Click for room
- Scale and click for room
- Resize and click for room

PR #1587, opened 2024-07-20
[eherbs]v2.0.9 code consolidation

Deysh:
mostly an update to determining inventory and the use_herb method. Combines the blood only option and removed that stand alone method as well. Also put the injury results in the escort method into a hash. No change in functionality (least none was intended.


OSXLich-Doug (approved):
LGTM overall

PR #1608, opened 2024-08-07
[echild.lic] v1.20.0 custom disabler

mrhoribu:
Allow for a custom disabler via `--disabler=<NUM>`


OSXLich-Doug:
Recommend converting to CharSettings so that the disabler (and potentially other settings) don't have to be re-entered as different characters are spun up.

PR #1610, opened 2024-08-09
[all][map.lic] Permit different settings per character

OSXLich-Doug:
Update `Settings` to `CharSettings` permitting different settings by character.


OSXLich-Doug:
I field questions about two or more characters on screen at the same time, and how `;map` stacks and one instance has to be moved.  Having said that - I suspect that is the area of greatest concern, the location.  Perhaps to test both theories, all settings should be global with the exception of position / location information.  Thoughts?

PR #1630, opened 2024-08-24
[sk.lic] v1.2.1 time_per fix

mrhoribu:
Uncertain if 10mins should be used as the baseline or not, but leaving as is for now.


ondreian:
I think it's 15 minutes for the floor, but also won't the next effects component update it?

PR #1637, opened 2024-09-03
[ecleanse.lic] v1.1.1 Wall of Thorns unpoison

mrhoribu:
Update to include WoT unpoisoning


OSXLich-Doug:
Will the same `.active?(/regex/)` work here as set for the `Status.thorned?` logic?

OSXLich-Doug (approved):
GTG

PR #1642, opened 2024-09-11
[bigshot]5.6.0 new claim process and bandit targeting

Deysh:
- new room claim process inspired by ;overwatch and Lich::Claim
- additional targeting for bandits


scripts/bigshot.lic
snperry85, line 1: `      confusion_debuff: lambda { @CONFUSION == true && Effects::Debuffs.active?("Confusion") },`

Looks like one more spot in the ready_to_rest?. I missed it and pulled off another murder. 😂

Deysh: Thanks, fixed

PR #1710, opened 2024-12-13
[SpellWindows.lic] initial creation

Nisugi:
Combines flextape with wrayth's spell active windows. Best of both worlds! A snappy Wrayth and up to date information.

Blocks active spells, buffs, debuffs, cooldowns, quickbar, and combat window feeds Uses Effects:: data to populate the spells, buffs, debuffs, and cooldowns data and sends it to the windows.

Choose which windows gets updated.
Can turn on combat window feed if you use that sucker.

I know Tysong prefers _respond, but _respond spams log.lic, and this only works with wrayth.

### DIFF
--- a/scripts/SpellWindows.lic
+++ b/scripts/SpellWindows.lic
@@ -1,0 +1,206 @@
+=begin
+  SpellWindow.lic
+  
+  Replaces Wrayths integrated spell active windows for spells, buffs, debuffs, cooldowns.
+  
+  The script blocks the xml feed for active spells, buffs, debuffs, cooldowns, quickbar, and combat.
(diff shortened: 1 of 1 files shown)


Nisugi:
Incorporated buff_tracker functionality of a customizable list of spells to list if they're not active.
Incorporated targetwindow functionality and expanded to list players and group members.

These are optional windows that are disabled by default. Incorporating them into a single script and reducing threads is a plus.

This video shows a character with one group member running to town and back. The top screen is a third character that shows the actual arrival time vs the experienced arrival time. Neither character have any scripts running to start. The second run is group leader with spellwindows active. Third run is both characters with spellwindows active.
https://www.youtube.com/watch?v=DoUZUshtOk4

PR #1773, opened 2025-02-07
fix(eloot.lic): v2.1.5 bugfix for hoarding inventory

ken-r-smith:
Fixes a bug with building the hoarding inventory when using a family vault as the container


Deysh:
Good catch. Thanks ken-r-smith

PR #1781, opened 2025-02-12
[Ewaggle] v2 convert to yaml and general code updates

Deysh:


OSXLich-Doug:
DId some testing - appears equivalent in functionality to prior version + no segfault on drag and drop.  Respects SK and performs casts to appropriate times / limits.  Good submit!

PR #1845, opened 2025-03-13
[SpellWindows.lic] correct Indef duration display

Nisugi:
added commands for missing spells
addall - adds all active spells
removeall - clears list

corrected duration display for items with indefinite duration


Nisugi:
Yeah I'm a dumbass.

PR #1866, opened 2025-03-24
[log_search.lic] initial creation

Nisugi:
Parses log.lic logs recording searches and notable finds.
Timestamps enable week by week breakdown but are not required.


Nisugi:
killtracker updated with more tracking to match what log_search pulls out.

alastir released killtracker .. not my favorite name but I dealt with the devil and people already expect the name hah.

PR #1931, opened 2025-04-27
Update bigshot.lic

ServerFailure:
This allows you to loot on an alt, but pick up leftovers on main.


Deysh:
Sure, I can do that.

PR #1954, opened 2025-05-08
Update kswole.lic

ADarkElf:
Updated to automate shield mind if know for characters that know kroderine soul or do not know kroderine soul.


ADarkElf (commented):
Fixing errors.

PR #2301, opened 2026-04-23
fix(log.lic): v1.1.0 timestamp reget lines and filter stream tags

MahtraDR:
## Summary
- Reget lines are now timestamped with `$login_time` when `--timestamp` is enabled, instead of being dumped raw without timestamps
- `pushStream`/`popStream` XML tags are filtered from reget output, matching the behavior of the main logging loop
- Version bumped to 1.1.0

## Context
Previously, `log.lic` used `file.puts(reget)` to dump the pre-launch buffer in bulk. This meant the first portion of every log file had no timestamps, even when `--timestamp` was enabled. Now each reget line is processed individually with the same filtering and timestamping as the main loop, using `$login_time` as the timestamp (since the lines arrived between login and script start).

## Test plan
- [x] Added comprehensive RSpec specs (`spec/log/log_spec.rb`) covering:
  - `Opts.parse` argument parsing (flags, values, comma-separated lists, positional commands)
  - Reget timestamping with `$login_time` when `--timestamp` is enabled
  - Reget lines written plain when timestamps are disabled
  - `pushStream`/`popStream` filtering from reget output
  - Skipping reget when login was more than 30 seconds ago
  - Reget marker comment presence
  - Log directory structure creation
  - Empty reget buffer handling
- [x] All 19 specs pass
- [x] Rubocop clean (0 offenses)

## Summary by CodeRabbit

* **New Features**
  * Enhanced reget payload processing to filter stream markers and support per-line timestamping in logs

* **Tests**
  * Comprehensive test coverage added for logging functionality, validating timestamp handling, payload filtering, and log file creation

* **Chores**
  * Version updated to 1.1.0 with corresponding changelog

### DIFF
--- a/scripts/log.lic
+++ b/scripts/log.lic
@@ -16,10 +16,13 @@
        author: elanthia-online
  contributors: Tillmen, Tysong, Xanlin
          game: gs
-      version: 1.0.0
+      version: 1.1.0
      required: Lich > 5.0.1
 
   changelog:
+    1.1.0 (2026-04-23):
+      * Timestamp reget lines with $login_time instead of dumping them raw
+      * Filter pushStream/popStream tags from reget output
     1.0.0 (2025-04-08):
       * Fix --rnum to better match only room titles
       * Add --exclude=288,u7199 support to exclude logging in rooms designated.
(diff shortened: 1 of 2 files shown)


MahtraDR:
Both @CodeRabbitai nitpicks addressed in 729014f:

1. **Stringly-typed `--lines`**: Added a comment clarifying that `parse_flag` returns strings and callers coerce via `.to_i` -- so the `eq('50000')` expectation is intentional.
2. **Brittle module extraction**: Added `raise` guards before `eval` to fail loudly if the regex returns `nil` or matches more than one `module GameLogger` block.

mrhoribu (approved):

PR #2302, opened 2026-04-23
fix(bigshot.lic): v5.12.5 improve final loot for solo hunting

mrhoribu:
Allows final loot to work for solo hunting as well. Which allows for looting dropped/left behind items that other players don't pick up, as well as being able to ensure when you "poof" a creature that it's loot that is dropped is also collected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended final-loot collection to work in solo mode, enabling comprehensive looting in the final room when playing solo or as a leader.
  * Updated user interface labels and tooltips to clarify final-loot behavior for both solo and leader modes.

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -941,214 +941,213 @@ end
 
 # Setup class for UI
 class Bigshot
-  if defined?(Gtk) || !$bigshot_headless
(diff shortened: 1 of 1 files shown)


mrhoribu:
@coderabbitai please remove all previous comments and review again

PR #2304, opened 2026-04-27
feat(ewander.lic): v1.1.0 priority matching and other fixes

Zedarius5:
Added 4 features:
- Target the NPC name that you're wandering for (if applicable)
- If multiple NPC name arguments, target the highest priority one (left to right) when any are found
- Checks for NPC's before wandering OUT of a room, in case they have walked in after you
- Ignores sorcerer animates

## Summary by CodeRabbit

* **New Features**
  * Extended `;ewander` command to auto-target matched NPCs with priority-based selection. NPCs are now automatically targeted in the order provided as arguments, with animated NPCs excluded from matching.

### DIFF
--- a/scripts/ewander.lic
+++ b/scripts/ewander.lic
@@ -10,7 +10,7 @@
   Use   ;ewander add              and
           ;ewander remove           to set up boundaries to keep you in your hunting area.
   Use   ;ewander delay <number>   to set the movement delay
-  Use   ;ewander <npc1> <npc2>    to stop only for specific npcs.
+  Use   ;ewander <npc1> <npc2>    to stop and target specific npcs (priority order, left to right).
   Use   ;ewander <pc>             to stop for another specific character.
   Use   ;ewander                  to stop for any targetable npcs.
 
@@ -18,9 +18,14 @@
      game: Gemstone
      tags: movement
  required: Lich >= 5.11.0
-  version: 1.0.0
+  version: 1.1.0
 
   changelog:
+    1.1.0 (2026-03-27):
+      - when given NPC name arguments, auto-target the matched creature before stopping
+      - NPC name arguments are treated as priority order (first arg = highest priority)
+      - check for matching NPCs before moving, catching creatures that enter mid-room
+      - skip player-controlled animates (names containing "animated")
     1.0.0 (2025-03-19):
       - initial fork of wander.lic
 =end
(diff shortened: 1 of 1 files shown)


scripts/ewander.lic
```diff
@@ -272,16 +277,41 @@ elsif script.vars.empty?
...
+    script.vars[1..-1].each do |name|
+      result = GameObj.npcs.find { |npc|
+        npc.status != 'dead' &&
+        npc.name =~ /#{name}/i &&
+        npc.name !~ /\banimated\b/i
+      }
+      break if result
+    end
+    result
+  }
```
mrhoribu, line 291: move `proc` outside of the main loop logic, move below the wander proc near top of script so that procs are defined before going into the script logic for easier management. Will eventually refactor the entire script into a module/class architecture away from procs. But for now, better to group the procs together.

scripts/ewander.lic
```diff
@@ -272,16 +277,41 @@ elsif script.vars.empty?
...
   loop {
+    # Check before moving — catches NPCs that entered while still in this room
+    if Claim.mine? && !GameObj.pcs.any? { |pc| pc.noun =~ /#{script.vars[1..-1].join('|')}/i }
+      if (found_npc = find_target.call)
+        fput "target ##{found_npc.id}"
+        waitrt?
+        break
+      end
+    end
```
mrhoribu, line 300: I concur with CodeRabbit's assessment here. Check it's suggested fix and work that in.

scripts/ewander.lic
```diff
@@ -272,16 +277,41 @@ elsif script.vars.empty?
...
       break
     elsif !Claim.mine?
       sleep CharSettings['delay']
-    elsif GameObj.npcs.any? { |npc| (npc.status != 'dead') and npc.name =~ /#{script.vars[1..-1].join('|')}/ }
-      break
     else
-      sleep CharSettings['delay']
+      # Find first live NPC matching arguments in priority order (first arg = highest priority)
+      if (found_npc = find_target.call)
+        fput "target ##{found_npc.id}"
```
mrhoribu, line 309: Doing a blind "target" for any npc is not going to be good. I would recommend doing something along the lines of:
```ruby
fput("target ##{found_npc.id}") if GameObj.targets.map(&:id).include?(found_npc.id)
```
This ensures it ONLY targets the npc if it's a targetable npc (hostile) as you do not want to target the town cleric when you're searching for it.

scripts/ewander.lic
mrhoribu: Only downside to doing the animated check your doing is if someone wants to find an animated something for whatever reason, also there are animated slushes. We get around this in GameObj.targets via:
```ruby
      def self.targets
        XMLData.current_target_ids.filter_map do |id|
          npc = @@npcs.find { |n| n.id == id }
          next unless npc
          next if npc.status.to_s =~ /dead|gone/i
          next if npc.name  =~ /^animated\b/i && npc.name !~ /^animated slush/i
          next if npc.noun  =~ /^(?:arm|appendage|claw|limb|pincer|tentacle)s?$|^(?:palpus|palpi)$/i &&
                  npc.name !~ /(?:amaranthine|ghostly|grizzled|ancient) kraken tentacle/i
          npc
        end
      end
```

mrhoribu (changes requested):

mrhoribu:
included the changes and more in #2307 
closing this as deprecated older code.
thanks for the priority matching changes

PR #2310, opened 2026-05-04
fix(autostart): v0.68 game-agnostic YAML autostarts, wayto overrides, and double-start guard

MahtraDR:
## Summary
- Adds `next if Script.running?(script_info[:name])` guard to the generic autostart loop to prevent double-starts
- Extracts YAML-based autostarts, `UserVars.autostart_scripts`, and `Map.apply_wayto_overrides` out from behind the `if XMLData.game =~ /^DR/` gate -- these features are now game-agnostic
- Adds a visible warning when a configured autostart script is not found on disk (typos, missing downloads)
- Simplifies the DR block to only dependency startup and the legacy first-run install path

## Problem
**Double-start bug:** The DR autostart loop (lines 138-163) guarded with `Script.running?`, but the generic Lich autostart loop (lines 197-227) did not. Scripts in both registries started twice on every login.

**DR-only gate:** YAML profile autostarts (`get_settings.autostarts`), `UserVars.autostart_scripts`, and `Map.apply_wayto_overrides` were gated behind `if XMLData.game =~ /^DR/` despite all underlying infrastructure (`get_settings`, `SetupFiles`, `Map`) being game-agnostic. GS characters could not use YAML profiles for autostarts or wayto overrides.

**Silent failures:** When a script configured in YAML profiles or `UserVars.autostart_scripts` did not exist on disk, it was silently skipped with no feedback to the user.

### `scripts/autostart.lic`
- New game-agnostic block (before the DR gate):
  - Merges `UserVars.autostart_scripts` + `get_settings.autostarts` (deduplicated)
  - Applies `Map.apply_wayto_overrides` if available
  - Starts each script with guards: skip dependency, skip obsolete, skip running, warn and skip if not found
  - Gated only on `respond_to?(:get_settings, true)` -- no game check
- DR block simplified: only starts `dependency` for runtime helpers + legacy first-run install path
- Generic Settings/CharSettings loop: added `Script.running?` guard to prevent double-starts

### `spec/autostart/autostart_spec.rb`
- 50 specs (up from 0), covering:
  - Generic autostart loop: happy path, running guard, cross-registry dedup, empty/nil/non-array inputs, skipped scripts, DR dependency removal, edge cases
  - YAML autostart loop: YAML autostarts, UserVars autostarts, merge/dedup, wayto overrides, skip guards, game-agnostic behavior (GSIV + DR)
  - GS character with no YAML profiles (nil autostarts, nil UserVars)
  - Adversarial inputs: nil entries, empty strings, duplicates, all running, none exist, both sources nil, dependency in both
  - Missing script warnings: per-script warning, no false warnings for running/dependency/existing scripts, mixed success+failure

## Test plan
- [ ] `rubocop scripts/autostart.lic spec/autostart/autostart_spec.rb` -- 0 offenses
- [ ] `rspec spec/autostart/autostart_spec.rb` -- 50 examples, 0 failures
- [ ] GS character with no YAML profile: no errors, no warnings, falls through to Settings/CharSettings loop
- [ ] GS character with YAML profile (`scripts/profiles/<CharName>-setup.yaml` with `autostarts:` array): starts configured scripts
- [ ] DR character: existing behavior preserved (UserVars + YAML merged, dependency started, legacy path works)
- [ ] Typo in YAML autostart list: visible warning at login

### DIFF
--- a/scripts/autostart.lic
+++ b/scripts/autostart.lic
@@ -134,33 +134,35 @@ if script.vars.empty?
     Lich::Util::Update.request("--announce")
   end
 
+  # YAML-based autostarts, UserVars autostarts, and wayto overrides.
+  # Game-agnostic -- requires modern Lich with get_settings support.
+  if respond_to?(:get_settings, true)
+    UserVars.autostart_scripts ||= []
+    all_autostarts = (UserVars.autostart_scripts.to_a +
+                      get_settings.autostarts.to_a).uniq
+
+    Map.apply_wayto_overrides if Map.respond_to?(:apply_wayto_overrides)
+
+    all_autostarts.each do |script_name|
+      next if script_name == 'dependency'
+      next if defined?(DR_OBSOLETE_SCRIPTS) && DR_OBSOLETE_SCRIPTS.include?(script_name)
+      next if Script.running?(script_name)
+
+      unless Script.exists?(script_name)
+        respond "\n--- autostart: '#{script_name}' not found, skipping\n\n"
+        next
+      end
+
+      start_script(script_name)
+    end
+    did_something = true unless all_autostarts.empty?
+  end
+
   if XMLData.game =~ /^DR/
-    if respond_to?(:get_settings, true) && respond_to?(:start_scripts_if_available, true)
-      # New path: core lich provides get_settings and start_scripts_if_available.
-      # Own the autostart loop directly instead of delegating to dependency.lic.
-
-      UserVars.autostart_scripts ||= []
-
-      # Resolve autostarts: UserVars + yaml 'autostarts' setting
-      dr_autostarts = (UserVars.autostart_scripts.to_a +
-                       get_settings.autostarts.to_a).uniq
-
-      # Apply personal map overrides if core provides it
-      Map.apply_wayto_overrides if Map.respond_to?(:apply_wayto_overrides)
-
+    if respond_to?(:start_scripts_if_available, true)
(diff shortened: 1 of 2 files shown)


scripts/autostart.lic
```diff
@@ -218,6 +220,8 @@ if script.vars.empty?
           end
           next
         else
+          next if Script.running?(script_info[:name])
+
           Script.start(script_info[:name], :args => script_info[:args])
```
mrhoribu, line 227: Script.start already prevents running a script if it's already running.Not sure if it's needed here?
```
>;e Script.start("ledger")
--- Lich: exec1 active.
--- Lich: ledger is already running (use ;force [scriptname] if desired).
--- Lich: exec1 has exited.
```

mrhoribu (commented):
Forgot to bump version of the script in the header, other than that and the one comment I left, looks good.

scripts/autostart.lic
MahtraDR: Good catch -- you're right that `Script.start` already guards against double-starts. The `running?` check here isn't for safety, it's to suppress the noisy `--- Lich: <script> is already running` warning that `Script.start` prints to the game window. Without it, any script that appears in both YAML autostarts and Settings/CharSettings autostarts would flash that warning on every login.

This is also the same pattern used in the YAML autostart loop earlier in this file (line ~144) -- keeping both call sites consistent.

mrhoribu: Roger, thanks for clarifying the purpose.

PR #2384, opened 2026-07-09
fix(resource.lic): v1.12.1 enable Battle Standard and Bloodstone Jewelry chance calculations

Lillymara:
## Summary
- Add Battle Standard (B1-B6) difficulty calculation to `resource.lic` chance command using the formula from gswiki: `75 + (tier * 75)`, giving absolute difficulties of 150, 225, 300, 375, 450, 525
- Fix Bloodstone Jewelry (J1-J5) to use the same correct formula instead of the previous placeholder values (150/75)
- Both use absolute difficulty (`item_difficulty = 0`), matching how Resistance and Covert Arts work
- Add Paladin to auto-bonus profession detection so `;resource chance <difficulty> B1` works without manually specifying a skill bonus
- Remove "(not implemented)" labels and update help text

## Notes
- Difficulty formula sourced from gswiki: both Battle Standard and Bloodstone Jewelry use `75 + (tier * 75)`
- The Paladin bonus formula was already fully implemented in the `bonus` method - only the `chance` method's profession regex was missing it


mrhoribu (changes requested):
Looks good, just need to bump the changelog and add a changelog entry

mrhoribu (changes requested):
looks good, but need to bump the header version and add a changelog entry.

mrhoribu (approved):

PR #2386, opened 2026-07-10
feat(cartographer.lic): v1.1.0 import cartographer + runtime fixes

Nisugi:
## What

Imports `cartographer.lic` (the mapdb release client, currently distributed only via the in-game `;repository`) into this repo so it has a proper git home, then applies five runtime fixes on top. The first commit is the script verbatim as distributed today, so each fix is reviewable as its own diff.

## Fixes

- **Windows: stringprocs extraction crashed** — `File.open` without a mode reads in text mode, which on Windows treats the first `0x1A` byte of the gzip stream as EOF (`Zlib::GzipReader: unexpected end of file`). Now opens `'rb'`.
- **StringProc file reads cached** — `evaluate_script` re-read its `.rb` file from disk on every call; route calculation invokes timeto procs as Dijkstra edge weights, so one pathfinding pass could trigger thousands of redundant reads. Contents are now cached per path (version dirs are immutable after extraction) and invalidated on base-path change. The `eval` still runs every call — results depend on live game state.
- **timeto failures no longer return a fake 0.2** — any error (unset base path, missing file, eval exception) silently fed pathfinding a made-up edge weight. Errors now return `nil`, which Map pathfinding treats as an unroutable edge (`next unless edge_weight`). Errors are still reported; wayto behavior (raise) is unchanged.
- **`Map.load_cartographer` sets the StringProc base path itself** — previously only the script's own download flow set it, so calling the Map extension directly loaded rooms whose StringProcs all failed.
- **GS-only guard** — the data dir is keyed on `XMLData.game` but mapdb releases contain GemStone rooms only; running on a DR character would install GS data into the DR map.

## Verification

- `ruby -c` and RuboCop (repo config) clean
- Standalone harness covering the `Cartographer` shim: happy path, cache-hit-after-file-deletion, fresh eval per call, cache invalidation on `set_base_path`, and nil-vs-raise for timeto/wayto across all three error branches (11 checks)
- Live end-to-end on Windows: `;cartographer --force` → download → extract → load of 36,332 rooms → built-in map self-test passing

### DIFF
--- a/scripts/cartographer.lic
+++ b/scripts/cartographer.lic
@@ -402,7 +402,7 @@ class CartographerManager
         # Extract stringprocs
         echo "Extracting StringProc files..."
         # system("cd '#{version_dir}' && tar -xzf stringprocs.tar.gz")
-        Gem::Package.new("").extract_tar_gz(File.open(temp_stringprocs), version_dir)
+        Gem::Package.new("").extract_tar_gz(File.open(temp_stringprocs, 'rb'), version_dir)
         if Dir.exist?(File.join(version_dir, "stringprocs"))
           File.delete(temp_stringprocs)
           echo "Successfully extracted StringProc files"
--- a/scripts/cartographer.lic
+++ b/scripts/cartographer.lic
@@ -717,9 +717,11 @@ end
 # Cartographer class for userland StringProc evaluation
 class Cartographer
   @@base_path ||= nil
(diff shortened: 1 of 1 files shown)


Nisugi:
Closing for now - we're assembling and testing the full contribution workflow (client + repo automation) before submitting the pieces. Will reopen when the set is ready.

PR #2387, opened 2026-07-13
fix(blackarts.lic): v1.3.6 sea water vial/flask validation logic

ChickenBeard:
## Summary
- Changed sea water vial/flask validation from AND to OR logic since you need one or the other but not both
- This fixes an issue where the script would fail if you had either a vial OR a flask, but required both

## Changes
- Bumped BlackArts to v1.3.6
- Updated validation check to accept either vial or flask for sea water
- Fixed typo in error message check ("Doe" → "Do")

### DIFF
--- a/scripts/BlackArts.lic
+++ b/scripts/BlackArts.lic
@@ -6,10 +6,12 @@
   contributors: Deysh, Tysong, Gob
           game: Gemstone
           tags: alchemy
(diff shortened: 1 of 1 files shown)


Deysh:
Sea water is collected with the flask.
Sea salt is extracted with a vial. 
Both are needed.

PR #2390, opened 2026-07-14
fix(bigshot): fix some beast belly edge cases

eightmelodies:
I've parked one of my characters on plane 4 for a bit now so I noticed a few issues with the built-in escape room stuff bigshot does:
1. If rest is triggered (eg., wounded) then bigshot will try to return to your rest room first, before `escape_rooms` gets called
2. Bigshot can try to loot inside the belly which causes eloot to loop forever because it doesn't handle the `can't do that while being digested` message. This happens every time if you have `final_loot` set.
3. On the way to the starting room, if you get snatched up and end up in the belly, bigshot will loop forever in the `goto` method.

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -17,6 +17,8 @@
 
   Version Control:
   Major_change.feature_addition.bugfix
+  v5.14.4
+    - fix some beast belly edge cases
   v5.14.3  (2026-07-13)
     - fix missing essence regex pattern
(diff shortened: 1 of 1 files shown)


scripts/bigshot.lic
mrhoribu: forgot to bump the version here

mrhoribu (changes requested):
looks good, just a minor header edit needed.

mrhoribu (approved):

PR #2395, opened 2026-07-27
fix(bigshot): v5.15.2 retarget when a higher priority creature arrives mid-fight

mrhoribu:
## Problem

With Priority Hunt enabled and an ordered target list, bigshot notices when a
higher priority creature enters the room but never switches to it. It keeps
fighting the current, lower priority target until that target dies or leaves, at
which point the arrival is picked up as an ordinary next-target selection rather
than a priority interrupt. Affects normal and quick hunting alike.

## Root cause

`priority(target)` memoized its answer against room composition
(`$current_room_npcs` vs `$room_npcs_last_check`, bigshot.lic:6977-6979), but the
answer is a function of `target`, which was not part of the key.

`priority()` has four callers, and only `do_hunt` (6166-6168) can actually
retarget. `attack_break` (6515) and `cmd` (3391) run far more often, so whichever
of them evaluated first consumed the one fresh computation and stamped the room as
"already evaluated." `do_hunt`'s check then short-circuited to true and never
called `find_target(nil)`. Verified both orderings against the real method body:

- attack_break first -> breaks the command loop, then do_hunt retarget = false
- do_hunt first       -> retargets correctly, then attack_break = false

Recomputing costs about 76 microseconds with a 23-entry target list and three
NPCs in the room, against a call rate bounded by roundtime, so there was nothing
worth caching.

Note for reviewers: the `x.id == y.id` comparison on the removed line did not
raise on `zip`'s nil padding, because Lich patches `NilClass#method_missing` to
return nil (lich-5 `lib/common/class_exts/nilclass.rb`). That nil tolerance was in
fact the only signal that detected an arrival at all, so hardening the comparison
without removing the guard would have made priority hunting worse.

A second, related bug surfaced during testing. `cmd_eachtarget` (3515-3525)
targets every creature in the room and calls `cmd` per creature, and `cmd`'s
priority gate (3391) aborted each of those calls. With a correct `priority()` that
became deterministic: every creature except the highest ranked one was targeted
and then skipped, which reads as target thrash in the log.

## Changes

- `priority()`: drop the room-composition cache and recompute per call.
- `priority()`: compare target list rank rather than identity. `GameObj.targets`
  is the game's target dropdown (`xmlparser.rb:464-478`, cleared and rebuilt
  wholesale), so the current target can be missing from it for a moment. Ranking
  off the list means only a creature that outranks the target may interrupt it,
  and a lower ranked creature can never win a check by default. New helpers
  `priority_matchers` and `priority_rank`; the debug line now reports
  `target_rank` and `best_rank`.
- `priority()`: anchor name and noun matching to match `sort_npcs` (6967), so it
  cannot nominate a creature `find_target` is unable to select.
- `priority()`: skip creatures whose ignored boons are already known.
  `invalid_target_with_boons` (6829) gains `assess = true`; passing false answers
  from `@BOON_CACHE` only, because `check_boons` sends `assess #id` to the game
  and a predicate called several times a second must not do that. The existing
  caller keeps the default and is unchanged.
- `cmd` gains `check_priority: true`; `cmd_eachtarget` passes false for its
  per-creature calls. The gate exists to abandon a single-target attack when
  something better appears, which is meaningless inside a sweep whose purpose is
  to hit everything. `attack_break` still evaluates between routine commands, so
  a genuine priority arrival interrupts at the next command boundary. All other
  `cmd` callers keep the gate.
- `do_hunt` (6166-6168) and the follower ATTACK loop (8280-8282): a retarget that
  finds no replacement keeps the current target instead of setting it to nil,
  which also removes a path that sent a bare `target #` to the game (6536).
- Follower loop: read `bs.PRIORITY` instead of `@PRIORITY`, a top-level instance
  variable that was never set, making follower priority retargeting dead code.
- Remove the unused `npc_room_check` snapshot thread (5730-5737), its two call
  sites (5968, 8192) and its two globals (487, 492). Nothing else in this repo
  referenced them.
- Header bump to v5.15.2 with changelog entry.

## Behavior changes

Reachable only where the priority list disagreed with what target selection can
pick:

- A partial list entry (for example `illoke` against `illoke mystic`) no longer
  claims a creature. Such creatures were never selectable, since `sort_npcs` has
  always matched anchored.
- When nothing that outranks the current target is present, bigshot keeps its
  target instead of dropping it or switching down the list.
- `eachtarget` routines now issue their command against every valid creature in
  the room while Priority Hunt is on, which is what they did before the cache
  stopped masking the gate.

## Testing

`spec/bigshot/priority_spec.rb`, 34 examples. The specs extract the real
`priority`, `priority_matchers`, `priority_rank`, `sort_npcs`, `find_target`,
`invalid_target_with_boons` and `cmd_eachtarget` bodies from `scripts/bigshot.lic`
along with the `do_hunt` and follower retarget blocks, and evaluate them against
stubs, so they fail if that code changes shape rather than drifting from it. Stubs
are namespaced inside the harness so the suite's real `GameObj` is untouched. No
`NilClass` patch: the fixed code must not depend on Lich's nil tolerance.

Coverage includes the reported room from the log (angargeist, disir and draugr
with the reporter's own list and creature ids) and the boon path, asserting that a
priority check never triggers an `assess`.

Targeted reverts confirm the specs bite:

- reverting the comparison to identity fails the two rank examples
- reverting the `cmd_eachtarget` call site fails the gate contract example
- reverting the whole `priority` method fails the spec's extraction guard

Verified locally: `ruby -c` clean, ASCII-only, CRLF preserved (8370 CRLF, zero
bare LF), rubocop reports only the pre-existing `cmd_jewel` offense that is
present on the unmodified file, and the spec subset that runs without the full
gem set passes 139 examples with no failures.

In game: the mid-fight priority arrival now retargets immediately, and
`eachtarget wandolier` hits every creature in the room again.

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -19,8 +19,10 @@
   Major_change.feature_addition.bugfix
   v5.15.2  (2026-07-27)
     - fix priority hunt not retargeting when a higher priority creature arrives mid-fight
+    - compare target list rank instead of identity so a lower ranked creature never wins a check
     - match priority targets with the same anchored name and noun rules used to select them
     - keep the current target when a priority retarget has no replacement to offer
+    - fix eachtarget commands skipping every creature except the highest ranked one
     - fix follower priority check reading an instance variable that was never set
     - remove the unused room npc snapshot thread and its globals
   v5.15.1  (2026-07-23)
@@ -3287,7 +3289,7 @@ class Bigshot
     end
   end
 
-  def cmd(command, npc = nil, stance_dance = true)
+  def cmd(command, npc = nil, stance_dance = true, check_priority: true)
     debug_msg(@DEBUG_COMMANDS, "cmd | command: #{command}")
 
     check_for_deaders_prone
@@ -3392,7 +3394,9 @@ class Bigshot
 
     # Target gone or not priority? Then return
     return unless valid_target?(npc)
-    return if @PRIORITY && !priority(npc)
+    # check_priority: false lets cmd_eachtarget sweep the room without the gate
+    # rejecting every creature that is not the highest ranked one
+    return if check_priority && @PRIORITY && !priority(npc)
 
     # Perform the command
     if (command =~ /^(incant)?\s?(\d+)\s?((?:open|closed)?\s?(?:cast|channel|evoke)?\s?(?:cast|channel|evoke)?\s?(?:open|closed)?\s?(?:acid|air|cold|earth|fire|lightning|steam|water)?)?.*$/i)
@@ -3523,7 +3527,7 @@ class Bigshot
     GameObj.targets.each { |target|
       next unless valid_target?(target)
       fput "target ##{target.id}" unless XMLData.current_target_id == target.id
-      cmd(command, target)
+      cmd(command, target, check_priority: false)
     }
     fput "target ##{current_target.id}" unless XMLData.current_target_id == current_target.id
   end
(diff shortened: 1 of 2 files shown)


mrhoribu:
@coderabbitai review

PR #2408, opened 2026-08-08
fix(egemhoarder): v2.0.1 correct regex to allow match of jar and bottle

mrjeckel:
Issue presented as the script hanging and terminating early when using *bottle*s


mrjeckel:
@coderabbitai review

PR #2432, opened 2026-08-26
fix(bigshot.lic): v5.16.0 looting races eloot on resume when both are paused

mrhoribu:
## Summary
- Fixes a rare timing bug in `looting_watch`: it only checked the *watched* loot script's pause state, never bigshot's own. When something (e.g. `ecleanse`) paused bigshot and the loot script back-to-back for a disarm-recovery cast, `looting_watch` would notice the loot script's pause and hand control back immediately - regardless of whether bigshot itself had also just been paused.
- That let bigshot advance one extra loot iteration and kill the loot script for the *next* corpse while the *current* one was independently resuming and finishing its own action, with no reliable ordering between the two. Could skip a loot/skin action or cut one off mid-command. Real session log reproducing it is in the linked work unit.
- Fix: block on bigshot's own pause at the top of the loop (the same idiom used elsewhere in the script) so it can't race ahead while paused.
- Landing as part of v5.16.0 alongside the rest of the stack rather than a separate release, and simplified the v5.16.0 changelog entries to plain language.
- Stacked on `docs/bigshot-yard-documentation` (elanthia-online/scripts#2417), the tip of the open bigshot PR stack (#2414 -> #2415 -> #2416 -> #2417 -> this).

## Test plan
- [x] New regression spec `spec/bigshot/looting_watch_spec.rb`, following the source-extraction-and-eval technique from `spec/bigshot/priority_spec.rb`. Uses a Mutex/ConditionVariable + Queue-based stub instead of sleep-based polling, so the race is deterministic:
  - blocks while bigshot is paused even though the loot script is also paused, and only returns once bigshot is unpaused
  - still breaks immediately on the loot script's own pause once bigshot itself was never paused (no behavior change for the common case)
  - kill/box-in-hand/rest-flag handling after the loop unaffected
- [x] Audited every other `Script.paused?`/`Script.running?` call site in bigshot.lic for the same bug shape - none found; this was the only raw polling loop watching another script's state without also checking its own
- [x] `bin/migrate` run first (gameobj-data fixtures)
- [x] `bundle exec rspec` - full suite passing
- [x] `RBENV_VERSION=4.0.5 rubocop -A` - clean

Generated with Claude Code

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -8,7 +8,7 @@
   contributors: SpiffyJr, Tillmen, Kalros, Hazado, Tysong, Athias, Falicor, Deysh, Nisugi
           game: Gemstone
           tags: hunting, bigshot, combat
-       version: 5.16.0
+       version: 5.16.1
       required: Lich >= 5.19.0
 
   Setup Instructions: https://gswiki.play.net/Script_Bigshot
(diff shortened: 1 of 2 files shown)


mrhoribu:
Superseded by elanthia-online/scripts#2433 - same commit, pushed directly to this repo instead of from my fork so it groups into the bigshot PR stack correctly.

PR #2443, opened 2026-09-02
fix(eherbs): refresh authoritative herb scans

therealatari:
## Summary

Prevent `eherbs` from making healing and stocking decisions from an older cached container snapshot.

- add an opt-in `refresh:` path to `Inventory.open_single_container`
- use it at the central authoritative herb-contents loader
- use it when an explicit `LOOK` has already established that a non-inventory herb container is closed
- retain the existing cached fast path for routine open requests

## Root cause

`open_single_container` treated `contents.is_a?(Array)` as proof that the contents were current. That only establishes that the container has been inspected before. If the published snapshot is older than the inventory state, an authoritative herb scan returns without issuing `OPEN` or `LOOK IN`, and `eherbs` can conclude that a needed herb is missing.

The refresh is centralized in `herb_container_contents_load` rather than added to every high-level healing or stocking action, avoiding duplicate scans while keeping the decision point authoritative.

This was also checked against the suspected Lich core regression. Lich's staged container refresh retains the last complete snapshot until the prompt commits a replacement by design, its focused core specs pass, and the relevant published-cache behavior is unchanged between Lich 5.19.1 and 5.20.1. This patch corrects the script's freshness assumption without weakening atomic `GameObj` publication.

## Testing

- Added a behavioral regression test that starts with an array-shaped stale cache and verifies the authoritative loader issues `OPEN` plus `LOOK IN`.
- Added a control test proving routine opens retain the cached fast path.
- `rspec spec/scripts/eherbs_spec.rb` — 2 examples, 0 failures
- `ruby -c scripts/eherbs.lic` — syntax OK
- `rubocop scripts/eherbs.lic spec/scripts/eherbs_spec.rb` — no offenses

### DIFF
--- a/scripts/eherbs.lic
+++ b/scripts/eherbs.lic
@@ -14,7 +14,10 @@
               game: Gemstone
               tags: healing, herbs
           requires: Lich >= 5.9.0
-           version: 2.2.0
+           version: 2.2.1
+  2.2.1 (2026-09-02)
+    - fix authoritative herb scans using a previously cached container snapshot instead
+      of refreshing the container before making healing and stocking decisions
   2.2.0 (2026-07-13)
     - add --splitblood option to count and stock major blood separately from acantha; off by default
     - fix check blood count to total every matching herb across all containers, not just the last item found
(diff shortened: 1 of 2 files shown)


therealatari:
Closing this after review of the Lich GameObj/XML container contract. I do not have a live EHerbs reproduction showing an open, tracked herb sack with array-shaped but stale contents. This change generalized from a separate ELoot disk-routing incident, and the added spec only verifies that a forced OPEN/LOOK occurs; it does not reproduce stale server-backed state. Because the blanket refresh also adds network round trips to the healing hot path, the change is not justified without a concrete repro. FLAG INVENTORY was ON in the affected sessions. Thanks to Tysong and the reviewer for catching the unsupported premise.

PR #2444, opened 2026-09-02
fix(ebounty): refresh authoritative bounty scans

therealatari:
## Summary

Prevent `ebounty` from making bounty decisions from older cached container snapshots.

- add an opt-in `refresh:` path to `EBounty.open_container`
- request authoritative scans before checking held heirlooms, forage herbs, gems, and skins
- preserve the cached fast path for routine open requests
- avoid scheduling a container for closure when it was already open before the forced scan

## Root cause

`open_container` treated `contents.is_a?(Array)` as proof that the contents were current. That only establishes that the container has been inspected before. With an older published snapshot, `ebounty` can skip `OPEN`/`LOOK IN`, miss an item already in the character's containers, and continue hunting unnecessarily.

Forced scans also need to distinguish "already open" from "opened by this script" so cleanup does not close a container the player had left open.

This was checked against the suspected Lich core regression. Lich's staged container refresh retains the last complete snapshot until the prompt commits a replacement by design, its focused core specs pass, and the relevant published-cache behavior is unchanged between Lich 5.19.1 and 5.20.1. This patch corrects the script's freshness assumption without weakening atomic `GameObj` publication.

## Testing

- Added a behavioral gem-bounty regression test: a stale empty snapshot is refreshed, the held gem is found, and the script prepares turn-in instead of hunting.
- Added coverage for already-open cleanup ownership and the routine cached fast path.
- `rspec spec/scripts/ebounty_spec.rb` — 3 examples, 0 failures
- `ruby -c scripts/ebounty.lic` — syntax OK
- `rubocop scripts/ebounty.lic spec/scripts/ebounty_spec.rb` — no offenses

### DIFF
--- a/scripts/ebounty.lic
+++ b/scripts/ebounty.lic
@@ -10,10 +10,13 @@
           game: Gemstone
           tags: bounty, adventure's guild, advguild, bounties
       required: Lich >= 5.19.0
-       version: 1.12.3
+       version: 1.12.4
 
   Version Control:
   Major_change.feature_addition.bugfix
+  v1.12.4 (2026-09-02)
+    - fix authoritative bounty inventory scans using previously cached container
+      snapshots when checking for heirlooms, herbs, gems, and skins
   v1.12.3 (2026-08-17)
     - fix rgambit usage when not hidden to just put away instead
     - fix to grab new bounty when available instead of waiting for should_hunt? to be true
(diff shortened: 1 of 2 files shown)


therealatari:
Closing this for the same evidentiary issue identified on #2443. I do not have a live EBounty reproduction showing array-shaped but stale contents for an open, tracked container. This was a defensive generalization from the separate ELoot disk-routing incident rather than a demonstrated EBounty failure, and its synthetic spec verifies the new forced-refresh behavior rather than the claimed bug. The Lich GameObj/XML contract already invalidates closed containers and publishes updates at prompt boundaries; without a contrary repro, the added round trips are unjustified. FLAG INVENTORY was ON in the affected sessions.

PR #2445, opened 2026-09-02
feat(eloot): manage unskinnable creature list

therealatari:
## Summary

- add `;eloot reset unskinnable` to clear the learned unskinnable-creature list without editing character YAML
- accept `clear` and `flush` as aliases
- add `;eloot remove unskinnable <creature name>` for exact, case-insensitive removal of one entry
- persist only when the list actually changes and document both commands in `;eloot help`

## Motivation

When a creature is incorrectly learned as unskinnable, users currently have to locate and edit `Lich5/data/GS4/<character>/eloot.yaml`, then reload eloot. These commands provide the same maintenance safely in game and mirror Bigshot's reset-style command.

## Testing

- 109 focused eloot examples pass
- Ruby syntax check passes
- RuboCop passes with no offenses on the changed files

The repository-wide suite could not be loaded locally because the available Ruby environment lacks the `rack` gem; the focused eloot suite and static checks are clean.

### DIFF
--- a/scripts/eloot.lic
+++ b/scripts/eloot.lic
@@ -15,9 +15,12 @@
               game: Gemstone
               tags: loot
           required: Lich >= 5.15.0
-           version: 2.11.4
+           version: 2.11.5
   Improvements:
   Major_change.feature_addition.bugfix
+  v2.11.5 (2026-09-02)
+    - feature: add commands to clear the automatically learned unskinnable creature
+      list or remove one creature from it without editing the character YAML file
   v2.11.4 (2026-08-31)
     - fix: bugfix for Vars.needed_reagents being a String
   v2.11.3 (2026-08-21)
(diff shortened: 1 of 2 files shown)


mrhoribu (commented):
Could we consolidate into a single CLI path?
```ruby
  when /^reset unskinnable\b/i
    ELoot.manage_unskinnable(Script.current.vars[3..].join(' '))
```

Shold then be able to change method to
```ruby
  def self.manage_unskinnable(creature)
    list = (ELoot.data.settings[:unskinnable] ||= [])

    if creature.empty?
      if list.empty?
        ELoot.msg(type: "info", text: " Unskinnable creature list is already empty.")
        return
      end

      removed = list.length
      list.clear
      ELoot.save_profile(silent: true)
      ELoot.msg(type: "info", text: " Cleared #{removed} #{removed == 1 ? 'creature' : 'creatures'} from the unskinnable list.")
    else
      creature = creature.strip

      removed = list.reject! { |entry| entry.to_s.casecmp?(creature) }
      unless removed
        ELoot.msg(type: "error", text: " \"#{creature}\" was not found in the unskinnable list.")
        return
      end

      ELoot.save_profile(silent: true)
      ELoot.msg(type: "info", text: " Removed \"#{creature}\" from the unskinnable list.")
    end
  end
```
and the help text:
```ruby
    rows << ["#{$lich_char}#{Script.current.name} reset unskinnable", "Clears the learned unskinnable creature list"]
    rows << ["#{$lich_char}#{Script.current.name} reset unskinnable <creature>", "Removes one creature from the unskinnable list"]
```

therealatari:
Superseded by #2442 so the disk-routing fixes and unskinnable-list management land together as one eLoot 2.11.5 update. The reviewed single-command interface is included there.

PR #2451, opened 2026-09-06
feat(bigshot.lic): v5.16.0 add thp<N> and empowered<N> command checks, fix buff<N> lookup

Nisugi:
Stacked on #2415 (base is its branch, so this reviews as ~50 lines and re-targets as the stack merges). Folds into the unreleased v5.16.0 rather than bumping the version. Two new command-check modifiers plus a fix to an existing one.

**thp<N>** - target HP percentage. `(thp20)` runs the command only when the target is at or below 20% of its max HP; `(!thp20)` only above. Reads `CreatureInstance#hp_percent` via a new `npc_hp_percent` helper (same shape as `npc_has_status?`): template max_hp minus Combat::Tracker-parsed damage. Skips when the npc has no CreatureInstance behind it.

Coverage note: with elanthia-online/lich-5#1539, 589/628 bundled templates carry a measured max_hp; without it most fall back to the guessed 400 and the percentage is only a rough heuristic. Damage-parsing coverage likewise improves with elanthia-online/lich-5#1559. Known drift is documented inline: regeneration/healing never credit damage back; damage with uncatalogued messaging is silently missed.

**empowered<N>** - Empowered buff strength guard. `(empowered30)` skips when an Empowered buff of +30 or more is already up, protecting a strong coup de grace endroll ((endroll - 100) / 5) from being overwritten by a weaker recast; `(!empowered30)` inverts. Freestanding - only reads Effects::Buffs. The existing `coupdegrace` modifier tests presence only and `buff<N>` tests duration only; neither compares magnitude.

**buff<N> lookup fix** - `command_check` reassigns `command` to the paren content before the buff branch runs, so `@COMMAND_BUFF_CHECKS[command]` looked up the whole modifier string ("barrage buff60") against bare-name keys ('barrage'). No digit-bearing paren content can ever equal a bare key, so buff_name was always nil and `buff<N>` was a silent no-op. Now resolves the key from the command being run (`original_command`), matching the documented usage ("buff60 with command 'barrage'"). This also makes combined lines like `(thp20 buff5 empowered30)` work as expected - modifiers AND together, first failing check skips the command.

Both new modifiers are added to the COMMAND_MODIFIER_REGEX whitelist; verified no capture collision with COMMAND_AMOUNT_REGEX. ruby -c clean.

### DIFF
--- a/scripts/bigshot.lic
+++ b/scripts/bigshot.lic
@@ -18,6 +18,7 @@
   Version Control:
   Major_change.feature_addition.bugfix
   v5.16.0  (2026-08-13)
+    - coup de grace eligibility gate, on by default: cmd_cmans re-tests the skill's real requirements at the moment of send - <= (rank*10)% HP incapacitated / (rank*5)% otherwise, 200 HP absolute cap (the binding limit on large creatures) - using the trained rank from CMan and live HP/statuses from the CreatureInstance. Holds the coup instead of wasting 20 stamina on a refusal; unknown targets (no creature data) pass through untouched. Incapacitating statuses in COUP_INCAP_STATUSES (prone confirmed live to qualify without a stun)
     - add empowered<N> command check: skip when an Empowered buff of +N or more is already active (!empowered<N> = skip unless). Guards a strong coup de grace endroll ((endroll - 100) / 5) from being overwritten by a weaker recast; the plain coupdegrace modifier only tests presence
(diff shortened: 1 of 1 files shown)


Nisugi:
Added in the second commit: a coup de grace eligibility gate, on by default. cmd_cmans re-tests the skill's actual requirements at the moment of send — at or below (rank×10)% HP when the target is incapacitated / (rank×5)% otherwise, with the 200 HP absolute cap (which is the binding limit on large creatures) — using the trained rank from CMan and live HP/statuses from the CreatureInstance. It holds the coup instead of spending 20 stamina on a guaranteed refusal; targets without creature data pass through untouched. Prone was confirmed in live testing to qualify as incapacitated without a stun; the full status set is in COUP_INCAP_STATUSES with a comment inviting corrections from further live testing.

PR #2456, opened 2026-09-09
feat(bigshot.lic): v5.17.0 expand Quick into bounded clear/watch/assist/trial/seek modes

therealatari:
## Summary

- expand Bigshot's existing `quick` path into bounded `clear`, `watch`, `assist`, `trial`, and `seek` modes that reuse normal profiles and routines
- add exact-owner supervision, runtime controls, profile-area admission, unknown-creature policy, native combat-outcome evidence, and ineffective-action limits
- add opt-in refuge outings that start safe, perform bounded work, restore equipment, and return before handoff
- keep ordinary Bigshot behavior unchanged when expanded Quick is not explicitly selected

## Scope and dependencies

This PR changes only Bigshot plus its Bigshot-specific documentation and tests, following maintainer guidance for script-scoped history and blame. Bigshot specs are organized under `spec/scripts/bigshot/`; smaller script suites remain flat.

It is intentionally stacked on #2433 (`fix/bigshot-looting-watch-pause-race`) and should be retargeted to `master` after the pending Bigshot stack merges.

Its optional integrations are independently proposed as:

- eLoot guarded room cleanup: #2457
- go2 one-trip observation preservation: #2458
- native Script execution guards: elanthia-online/lich-5#1575
- native combat observation provenance: elanthia-online/lich-5#1576
- static-only Map routing: elanthia-online/lich-5#1577
- cooperative Spell wait integration: prepared separately and queued behind #1575

Expanded Quick must not ship as compatible with an unmodified Lich 5.21 installation. Once the required Lich APIs receive a release version, the `required` metadata and documentation should be updated before merge.

## Safety

Expanded Quick fails closed on authority loss, target/room/session drift, stale or ambiguous combat evidence, unowned child scripts, unsupported routines, and exhausted action/time budgets. Standalone Bigshot behavior remains unchanged unless a recognized new Quick mode is explicitly used.

## Maintainer review revision

- preserve historical permissive Quick arguments such as `quick bounty`, `quick giant rat`, `quick solo`, and `quick once bounty`
- add opt-in lifecycle, seek, retreat, refuge, command, and error diagnostics through Bigshot's existing debug categories
- latch swing-wait interruptions through the active guard and retain bounded execution-error metadata
- make seek policy denials return `false` consistently and handle bare interrupted adapter outcomes safely
- surface Quick validation failures even when another setup tab is visible
- add plain-language hover help to every Quick preset and named-trial field and control
- keep the legacy tether regression runnable without unreleased native Lich APIs
- update YARD contracts, changelog language/date, and the stale `quick_area` plan reference

## Verification

- isolated Bigshot suite: 783 examples, 0 failures; 80 native-integration/real-GTK examples explicitly gated
- combined native Lich execution-guard/combat-provenance suite: 783 examples, 0 failures; only 5 real-GTK examples gated
- real GTK/Xvfb suite: 5 examples, 0 failures, including 1080x800 and 1080x720 rendering, save/discard behavior, and native Settings persistence
- all modified Ruby/LIC files pass RuboCop, Ruby syntax validation, and whitespace checks
- player-authorized live acceptance: 28 rooms outbound, four attributed/game-confirmed kills, 28 rooms back to the configured refuge, original hands/standing/full health restored, owners released, and no alerts
- separate player-authorized ordinary-stop run returned to the same refuge with equipment and ownership restored

The live checks validate those exact no-loot runs, not every route, creature, profile, assist/trial combination, or live room-scoped eLoot path. Those limits remain in the committed smoke-test documentation. Dates in that record use ICT and include the corresponding UTC date.


therealatari:
@coderabbitai review

therealatari:
Addressed the actionable findings from the earlier combined diff in `fba781fb`:

- documented where refuge/supervision launch values are injected
- clarified the live-acceptance timezone (`2026-09-10 ICT` / `2026-09-09 UTC`)
- added Markdown fence languages
- forwarded guard wrapper arguments
- made the provenance copy test exercise a mutable source string
- validated both extracted safety methods before concatenation

The eLoot empty-corpse-list finding moved with that script and is fixed in elanthia-online/scripts#2457. Horibu's requested test organization is in `0a9c9714` under `spec/scripts/bigshot/`.

@coderabbitai review

mrhoribu (changes requested):
# PR Review: feat(bigshot.lic): expand Quick combat for bounded encounters (elanthia-online/scripts#2456)

**Reviewed:** 61 files, +16,758/−217 at `0a9c971` on 2026-09-09. Based on: full local clone (worktree at PR head), a local `elanthia-online/lich-5` checkout for API verification, plus executed RuboCop, `ruby -c`, and the full RSpec suite.
**Verdict:** Request changes — 1 blocker, 2 major, 10 minor, 2 open questions. The blocker the author already identified; neither major is.
**Summary:** This turns Bigshot's one-word `quick` into a five-mode bounded-encounter subsystem with supervision, guarded execution, and refuge outings. The engineering is unusually careful — nearly every path fails closed, and I could not break the new code from the outside. Two things hold it back. The new CLI gate narrows what `;bigshot quick …` accepts, so four historically valid invocations now hard-refuse. And the new subsystem ships with none of the instrumentation the rest of the script relies on — zero `debug_msg` calls in 4,254 new lines against the legacy code's 167, and YARD coverage down from 100% to 62% on a file that was fully documented two commits below the merge base.

---

## Blockers

### 1. The headline feature cannot run on any released Lich, and `required:` cannot express that — `scripts/bigshot.lic:11-14` (header block)

```ruby
       version: 5.17.0
      required: Lich >= 5.21.0
     companion: Lich PR #1575 until its execution-guard APIs receive a release version
    integrates: eLoot PR #2457 and go2 PR #2458
```

Expanded Quick depends on Lich APIs that do not exist in any published build. I checked each one against a current `lich-5` checkout (`LICH_VERSION = '5.20.1'`, `lib/version.rb:3`): `with_execution_guard`, `check_execution_guard!`, `execution_guard_active?`, and `execution_sleep` (from #1575), plus `Game.current_ingress_time` and `Combat::Tracker.observation_context` (from #1576), all return zero hits across `lib/`. (`Game.thread`, `Combat::Parser.parse_attack`, `Map.ids_from_uid`, `Experience.percent_fxp`, `Settings.root_proxy_for` and `CharSettings.active_scope` — the other new cross-repo calls — do all exist, so the gap is precisely the two unmerged PRs.) The three Lich PRs named in the body — elanthia-online/lich-5#1575, #1576, #1577 — are all still `OPEN`, and the newest Lich release is v5.20.1 (2026-08-20).

The runtime behaviour is correct and graceful: `admit_owner!` at `scripts/bigshot.lic:2198` checks `%i[with_execution_guard execution_guard_active? execution_sleep].all?` and raises `'Quick Combat requires the companion Lich script execution guard'`, which the entry point at `scripts/bigshot.lic:13832` catches and echoes. Nobody gets a stack trace. But a user on a stock install sees a version header advertising `Lich >= 5.21.0`, types `;bigshot quick clear`, and is told they are missing something the header never mentioned as a requirement.

The PR body states this ("Expanded Quick must not ship as compatible with an unmodified Lich 5.21 installation… the `required` metadata and documentation should be updated before merge"), so this is confirmation rather than news: **do not merge until #1575 and #1576 land in a Lich release and `required:` names that version.** The `companion:`/`integrates:` header keys themselves are harmless — `bin/repo:106` uploads the whole `=begin…=end` block verbatim without key validation, and the two consuming regexes (`version: (\d+\.\d+\.\d+)` and `required: Lich >= (\d+\.\d+\.\d+)`) are unaffected.

*Out of scope, but worth stating once for the stack:* `required: Lich >= 5.21.0` arrives from the base branch (commit `a07b2c25`, "fix: update required Lich to 5.21.0"), not from this PR. Until 5.21.0 ships, that line makes **all** of Bigshot exit at `scripts/bigshot.lic:443` for every user, quick or not. That is a #2433 concern, not a #2456 one.

---

## Major

### 2. Four previously working `;bigshot quick …` invocations now hard-refuse — `scripts/bigshot.lic:428-429`

```ruby
quick_words = Script.current.vars.drop(2).map { |word| word.to_s.downcase }
quick_extended = Script.current.vars[1].to_s.casecmp('quick').zero? && ![[], ['once'], ['single']].include?(quick_words)
```

`LEGACY_ARGUMENTS` (`scripts/bigshot.lic:662`) defines the legacy surface as exactly three forms: bare, `once`, `single`. Anything else after `quick` is routed to `QuickRequest.parse`, which rejects unknown first words. But the legacy dispatcher this replaces never had a fixed vocabulary — `scripts/bigshot.lic:13845` accepts *any* tail:

```ruby
elsif (Script.current.vars[1].nil? || Script.current.vars[1] =~ /solo|bounty|quick|single|once/i)
  $bigshot_quick = true if Script.current.vars[1] =~ /quick/i
  bs = Bigshot.new(Script.current.vars)
```

`@BOUNTY_MODE` is set from `options.any? { |var| var =~ /bounty/i }` — a scan of the *whole* argument list — and `@TRACKING_CREATURE` is `options[0].gsub(/(solo|bounty|quick|single|head|tail)\s*/i, '').strip` (`scripts/bigshot.lic:6982`), a regex that strips `quick` precisely so it can co-occur with a creature name.

I confirmed the change by evaluating the real `QuickRequest.parse` from this commit against Lich's actual argument tokenizer (`lib/common/script.rb:1736-1739`):

| invocation | v5.16.0 | this PR |
|---|---|---|
| `;bigshot quick` / `quick once` / `quick single` | quick | quick (unchanged) |
| `;bigshot bounty quick` | hunt+bounty | hunt+bounty (unchanged) |
| **`;bigshot quick bounty`** | **quick+bounty** | `Unknown quick command: bounty` |
| **`;bigshot quick giant rat`** | **quick+track** | `Unknown quick command: giant` |
| **`;bigshot quick solo`** | **quick** | `Unknown quick command: solo` |
| **`;bigshot quick once bounty`** | **quick+bounty+track** | `Unknown quick command: once` |

Two of these have no workaround, because `$bigshot_quick` is set only from `vars[1]`:

- **`;bigshot quick bounty` was the only way to run quick-mode bounty hunting.** Reversing it to `;bigshot bounty quick` sets `@BOUNTY_MODE` but leaves `$bigshot_quick` false, so it silently hunts normally instead. That reversal is *also* the exact example the script's own help prints — "Modes can be combined, e.g. `;bigshot bounty quick`" (`scripts/bigshot.lic:6605` and `:6628`) — so the documented combination has never actually worked, and the one that did now errors.
- **`;bigshot quick <creature>` was the only way to combine quick mode with Ranger tracking** (`@TRACKING_CREATURE`, consumed at `scripts/bigshot.lic:13428` and `:13501`). `;bigshot giant rat` alone falls through to `else` → `"Unknown option: giant"` (`scripts/bigshot.lic:14231`), so dropping `quick` is not an option.

The comment above the gate says the strictness is deliberate — "even a typo must never fall through to legacy hunting… its historical permissive tail also accepted typos" — and that is a fair goal. The issue is that the compatibility set was derived from the new design's vocabulary rather than from what the old dispatcher actually accepted, so it catches real usage along with typos.

The narrow fix is to invert the default: treat the tail as extended Quick only when it *positively* looks like Quick syntax, and let everything else fall through as before.

```ruby
QUICK_VERBS = (QuickRequest::RUN_COMMANDS + QuickRequest::CONTROL_COMMANDS + QuickRequest::TARGET_CONTROLS)
quick_extended = Script.current.vars[1].to_s.casecmp('quick').zero? &&
                 !quick_words.empty? &&
                 (QUICK_VERBS.include?(quick_words.first) || quick_words.first.start_with?('--'))
```

I checked this against the same harness. All four regressed forms return to legacy, and every real Quick form still reaches the parser:

| invocation | current gate | proposed gate |
|---|---|---|
| `quick` / `quick once` / `quick single` | legacy | legacy |
| `quick bounty` / `quick solo` / `quick giant rat` / `quick once bounty` | Quick → refused | **legacy** |
| `quick clear` / `status` / `stop` / `engage 12` | Quick | Quick |
| `quick trial burst --target 5` | Quick | Quick |
| `quick seek --area profile` | Quick | Quick |
| `quick --profile foo` | Quick → refused | Quick → refused (correctly — no mode given) |

It does give up the typo guard for `quick claer`, which would resume being read as a tracking creature — if you would rather keep that, an explicit allow-list of the legacy modifier words (`solo`, `bounty`, `once`, `single`) plus a free-text single-token tail gets most of both. Either way the choice deserves to be explicit rather than a side effect. `confirmed` — demonstrated by executing this commit's own parser.

*Pre-existing, not caused by this PR, but adjacent enough to be worth a free fix if you touch the gate:* `once` is missing from the `@TRACKING_CREATURE` strip regex at `scripts/bigshot.lic:6982`, so `;bigshot quick once` leaves `@TRACKING_CREATURE = "once"` and a Ranger will run `track once` at `scripts/bigshot.lic:13501`.

---

### 3. 4,254 new lines ship with zero troubleshooting output, in a script built around it — `scripts/bigshot.lic`

Bigshot's existing debugging story is one of its better features: `debug_msg(type, msg)` (`scripts/bigshot.lic:6780`) gated on `$bigshot_debug`, with four independently-toggled categories, called **167 times** across the legacy code:

| category | calls in legacy code | calls in new Quick code |
|---|---|---|
| `@DEBUG_COMBAT` | 45 | 0 |
| `@DEBUG_COMMANDS` | 66 | 0 |
| `@DEBUG_STATUS` | 23 | 0 |
| `@DEBUG_SYSTEM` | 31 | 0 |
| **total** | **167** | **0** |

Not one `debug_msg` call appears anywhere in the 4,254 lines this PR adds to `bigshot.lic`. The plumbing was deliberately preserved — `scripts/bigshot.lic:2351` computes `$bigshot_debug` from the engine's four debug ivars specifically so it survives the `reset.call` — but nothing was added for it to switch on. And `scripts/bigshot.lic:6932` forces `@DEBUG_FILE = false` on the quick path, so the file logger is unavailable to Quick runs too. A player who enables every debug flag and runs `;bigshot quick clear` gets exactly the same output as one who enables none.

Part of this is structural and fair: `EncounterController`, `QuickRun`, `QuickGuard` and friends are separate classes, not `Bigshot` instances, so they cannot call `debug_msg` without being handed a logger. But 25 of the 27 new `quick_*` methods live on `Bigshot` itself or in `QuickExecution`, which is mixed into it — `quick_execute`, `quick_observation`, `quick_safety_reason`, `quick_incant_scope`, `quick_settle_outcome_evidence`, `wait_for_swing`'s native branch. All 25 can call `debug_msg`; none do. (Only `BSAreaRooms#quick_status` and `#quick_steps` genuinely cannot.)

What makes this more than a style point is that **the instrumentation already exists and is already paid for; it is simply never shown to anyone.** `EncounterController#record_observation` maintains a 100-entry ring of every command, outcome, send count, reason and evidence reason, and `status` rebuilds the whole thing on every tick. I measured that copy: ~0.45 ms and ~1,541 object allocations per call, at roughly 10 Hz. It is spec-tested in detail (`spec/scripts/bigshot/quick_reporting_spec.rb` asserts ring length, truncation, sequence ranges and frozenness). And no runtime path reads it. `observations`, `observations_total`, `observations_dropped` and `observation_sequence_range` have **zero** consumers outside specs; `search`, `area`, `refuge`, `loot_recovery`, `escape_reason` and `work_result` are written into every snapshot and never rendered either.

Across an entire successful Quick run a player sees four strings: the start banner (`:2395`), `Quick Combat held: <reason>` when it holds (`:1736`), the final result (`:2398`), and — only if they ask — the `quick status` line (`:1880`), which reports state, target, action count and reason and nothing else. The remaining messages are control acknowledgements and input refusals.

This matters most precisely where the risk is highest. 81 of 779 examples cannot run in CI (see the open questions), so for the native execution-guard, attack-provenance, refuge-travel and retreat paths, **field reports are the primary validation channel** — and those reports will say `Quick Combat held: execution_error` with nothing behind it. Finding #5 below is the sharpest instance of the same gap.

The fix is small because the data is already there: emit the observation ring and the `search`/`area`/`refuge` sections under `debug_msg(@DEBUG_SYSTEM, …)`, add `debug_msg(@DEBUG_COMMANDS, …)` at the `quick_execute`/`quick_incant_scope` boundaries the way `cmd` already does at `:7621`, and consider extending `quick status` to print the last few observations. Passing the engine (or a small logger lambda) into `QuickRun`/`EncounterController` would let the pure classes participate too, without breaking their I/O independence.

---

## Minor

### 4. `wait_for_swing` raises the guard exception without latching it — `scripts/bigshot.lic:10634` and `:10639`

```ruby
raise QuickGuard::Interrupted, 'swing_target_missing' unless target && target.id.to_s.match?(/\A[1-9]\d*\z/)
```

`QuickGuard#interrupt!` (`scripts/bigshot.lic:1653`) exists specifically so an interruption is sticky: *"Legacy helpers may rescue an interruption, but a subsequent checkpoint must never revive this routine's permission."* These two sites raise the exception class directly, so `@interruption` is never set and a later `checkpoint!` would pass. `@quick_guard` is in scope and already used three lines down for `@quick_guard.deadline`, so `@quick_guard.interrupt!('swing_target_missing')` is a drop-in.

I traced the current call path — `cmd` → `wait_for_swing` → `quick_execution_scope` → `quick_execute` → `QuickRun#dispatch` — and nothing between them rescues `StandardError`, so today this behaves identically. Every other broad rescue in the class (`Group#member_online:4372`, `Group#has_bounty?:4484`, `trained_coup_rank:12127`, the DRb event loop at `:14219`) is off this path. So this is a latent inconsistency with the class's own stated invariant rather than a live bug — worth fixing because the invariant is exactly a defence against future callers. `confirmed` as an inconsistency; no current exploit path.

### 5. `dispatch` discards the exception it catches — `scripts/bigshot.lic:3739-3740`

```ruby
rescue StandardError
  { outcome: :interrupted, sends: @active_guard ? @active_guard.sends : 0, reason: 'execution_error' }
```

Any unexpected failure anywhere in the 4,000-line execution path collapses to the string `execution_error`. The player sees `Quick Combat held: execution_error` (from `quick_run_loop`, `scripts/bigshot.lic:1736`) with no class, message, or location, and `$bigshot_debug` does not change that. `QuickSeek` gets this right at `scripts/bigshot.lic:3111` — it captures `{ class: error.class.name, location: File.basename(...) }` into `@error` and surfaces it through `status`. Doing the same here (or a `debug_msg(@DEBUG_SYSTEM, …)`) would make field reports actionable for a subsystem that has no live users yet to shake bugs out.

### 6. A bare `:interrupted` from a dispatch adapter raises instead of interrupting — `scripts/bigshot.lic:1390`

```ruby
if outcome == :interrupted
  return status if terminal?
  reason = result[:reason].to_s
```

`tick` explicitly supports two dispatch return shapes — a Hash, or a bare outcome symbol (`scripts/bigshot.lic:1381-1383`, `else outcome = result`). On the symbol path `result` is a `Symbol`, and `Symbol#[]` is an index operation, so `:interrupted[:reason]` raises `TypeError: no implicit conversion of Symbol into Integer`. The method-level `rescue StandardError` at `:1414` then reports it as a dispatch failure rather than the interruption it was.

I reproduced it against the class extracted from this commit:

```
bare :sent (spec-exercised)           state=running  reason=nil
bare :ineffective (spec-exercised)    state=running  reason=nil
bare :interrupted (NOT exercised)     state=held     reason="dispatch_error: no implicit conversion of Symbol into Integer"
Hash interrupted (production shape)   state=running  reason=nil
```

Not player-reachable today: `QuickRun#dispatch` always returns a Hash on every path, including its own rescues. But the bare-symbol contract is real and is exercised by `spec/scripts/bigshot/encounter_spec.rb:24` and `:213` (`:sent`, `:ineffective`) — `:interrupted` is simply the one symbol never tested, and it happens to be the only one that reads `[:reason]`. Guarding the read (`reason = result.is_a?(Hash) ? result[:reason].to_s : ''`) matches how the two lines directly above it already handle the same ambiguity, and a spec case with a bare `:interrupted` would keep it honest. Lead from a local CodeRabbit pass; mechanism and repro verified here.

### 7. A validation error on the Quick tab silently blocks saving unrelated hunting settings — `scripts/bigshot.lic:5701-5707`

```ruby
def on_close_clicked
  return if @encounter_settings && !save_encounter_form
  return if @encounter_settings && !save_quick_trial(allow_empty: true)

  pre_save
  UserVars.op = @settings
```

`save_encounter_form` returns `false` on any `ArgumentError` from `EncounterSettings#save` and writes the reason to `@encounter_message`, a label inside the Quick Combat tab. `on_close_clicked` then returns before `pre_save` and `UserVars.op = @settings`, so the user's **hunting profile** edits are not written and the window does not close. Nothing switches the notebook to the offending tab — there is no `set_current_page` anywhere in the file — so if the user is looking at any other tab, clicking Close appears to do nothing at all, with no message anywhere they can see.

Concretely: open Setup, change a hunting setting, type `abc` into "Maximum actions per target" on the Quick tab (or clear the preset name), switch back to the combat tab, click Close. Nothing happens, repeatedly, with no feedback.

This matters more than it looks because `build_encounter_tab` runs for **every** user who opens Setup, not just Quick users — so this coupling is on by default. Nothing is lost (the window stays open, so the settings are still there), which is why this is Minor rather than Major, but "the Close button stopped working and I don't know why" is a support burden.

Two small fixes, either sufficient: switch the notebook to the Quick tab before returning, and/or surface the reason through `Lich::Messaging.msg` the way the success path already does. Independently, consider whether a bad Quick preset should block saving hunting settings at all — staging the Quick save and letting the rest through would decouple them.

### 8. Two controller ivars are never initialised — `scripts/bigshot.lic:1038-1076`

`EncounterController#initialize` scrupulously initialises every ivar it uses, including `@room_engaged`, `@engagement_roster`, and `@active_target` — but not `@encounter_started` (read at `:1103` and `:1316`) or `@retreat_attempted` (read at `:1285`). Both rely on the nil default, which works, and both are correct under `-w`-less execution. Given how deliberate the rest of the constructor is, the omission reads as accidental rather than intentional, and it costs one line each to make the initial state explicit.

### 9. YARD coverage drops from 100% to 62% on a file documented two commits ago — `scripts/bigshot.lic`

The base of this PR is *fully* documented, and recently and deliberately so: commit `1885fa0d docs(bigshot.lic): v5.16.0 add YARD documentation throughout` sits two commits below the merge base on `fix/bigshot-looting-watch-pause-race`, and its own message records "296 of 296 methods now carry `@param`/`@return`". I measured both revisions:

| revision | methods | with `@param`/`@return`/`@raise`/`@yield` |
|---|---|---|
| base `273fa778` | 299 | 299 (**100.0%**) |
| head `0a9c971` | 496 | 307 (**61.9%**) |

That is **+197 methods and +8 tagged — 4.1% of the new methods carry YARD tags.** Only 31% have any comment at all.

The gap is not a lack of care about comments: the new code is full of good *design-rationale* prose ("Evidence must identify both a current group member and an exact current hostile creature id", "Count before attempting transport: an exception is not proof no command reached the game"). Those are valuable and should stay. What is missing is the *interface* half — what goes in, what comes back, what it raises. And the author clearly knows the convention, because a handful of new methods do have it: `BSAreaRooms.for_quick` (`:4177`), `quick_steps` (`:4211`), `QuickSeek#call` (`:3067`), `QuickGuard#checkpoint!` (`:1620`). It is applied inconsistently rather than unknown.

The undocumented set includes essentially every entry point a future maintainer would start from:

```
686   QuickRequest.parse           prose comment, no tags
833   EncounterSettings.normalize  no comment, no tags
1015  EncounterPolicy#eligible?    no comment, no tags
1268  EncounterController#tick     no comment, no tags
1901  Bigshot#quick_execute        no comment, no tags
2194  QuickStartup.admit_owner!    no comment, no tags
3316  QuickRun#tick                no comment, no tags
3336  QuickRun#close               no comment, no tags
```

The sharpest case is the return contracts. The new region defines **7 different `call` methods and 4 different `result` methods**, each returning a differently-shaped Hash, and the controller branches on those exact shapes — `result[:outcome] == :retreated`, `cleanup[:outcome] == :complete`, `{ outcome: :pending }`, `{ outcome: :found }`, `{ outcome: :interrupted, reason: … }`. Six of the seven `call` methods and all four `result` methods say nothing about their return shape. `QuickSeek#call` (`:3067`) is the lone exception — `@param`, `@yieldreturn` and `@return [Hash] moved/found or explicit interrupted outcome (not combat proof)` — and it is the model the other ten should follow. This ambiguity is not hypothetical: it is what produced finding #6, where `EncounterController#tick` accepts two different dispatch return shapes and the undocumented one is broken.

A `@return [Hash]` line on each adapter's `call` and `result`, in the shape `QuickSeek#call` already uses, would be worth more than the rest of the YARD pass combined.

### 10. Changelog is dated tomorrow — `scripts/bigshot.lic:22`

`v5.17.0  (2026-09-10)` while today is 2026-09-09 and every other entry (`v5.16.0  (2026-08-13)`) uses a plain repo-frame date. The docs were correctly amended to `2026-09-10 ICT (2026-09-09 UTC)` after CodeRabbit's earlier note, but the header entry still carries the bare ICT date, so it reads as a future date against its neighbours.

### 11. Changelog entries are pitched at reviewers, not players — `scripts/bigshot.lic:22-26`

This repo's changelog entries are written for a player skimming what changed, and `v5.16.0` right below is a good model ("creature tracking is more accurate right after a kill or a room change, so targeting and looting are less likely to grab a stale target"). "added supervised controls, exact-target engagement, unknown-creature policy, and observed ineffective-action limits" and "added guarded room-scoped eLoot integration without allowing selling, child-script travel, or unrelated cleanup" are closer to a design summary. That detail is valuable — it just belongs in the PR description and `docs/`, which already carry it.

### 12. The plan document contradicts the contract document it ships alongside — `docs/bigshot-encounters-plan.md:109`

```
no movement authority. LAB uses the existing opt-in controller manifest and
`quick_area` proof, requires movement/combat ownership, and adds no listener or
```

Two other documents in this same PR say the opposite. `docs/bigshot-encounter-lab-contract.md:36-37`: *"Registered Quick tests require a `quick_refuge` handoff with an explicit refuge and return reserve. Legacy `quick_area` entries remain readable for migration"* — i.e. readable but not launchable. `docs/bigshot-quick-smoke-test.md:233-234` is blunter: *"The earlier field-only test plan is superseded: `quick_area` cannot launch an agent test. Register `quick_refuge`…"*. The plan was written against the earlier design and this line was not carried forward.

Surfaced by a local CodeRabbit pass; I verified it against all three files at head. It is a one-line fix. Worth doing rather than ignoring, because `bigshot-encounter-lab-contract.md` is the document a future maintainer will trust, and having the plan disagree with it about which handoff can launch a test is the kind of thing that costs an hour later.

### 13. PR title is missing the version number

The repo convention is `type(script): vX.Y.Z description`, consistent across merged history — including the previous Bigshot PR, #2451 `feat(bigshot.lic): v5.16.0 add thp<N> and empowered<N> command checks, fix buff<N> lookup`. This one reads `feat(bigshot.lic): expand Quick combat for bounded encounters`. Suggest `feat(bigshot.lic): v5.17.0 expand Quick into bounded clear/watch/assist/trial/seek modes`.

---

## Open questions

- **`scripts/bigshot.lic:3087-3097` — `QuickSeek`'s guard policy raises through `with_execution_guard`, while its three siblings return `false`.** `quick_execution_scope:2148`, `QuickRetreat130:2595`, and `QuickRetreatWalk:2856` all wrap their policy body in `rescue Interrupted → false`, converting an interruption into a clean denial. The seek movement lambda calls `observe!` and `fail!` with no rescue, so a `QuickGuard::Interrupted` unwinds out of the policy callback into Lich's guard machinery. `QuickSeek#call`'s outer `rescue StandardError` still turns it into `{ outcome: :interrupted }`, so the *script* recovers. **Settles it:** in elanthia-online/lich-5#1575, what does `with_execution_guard` do when the policy callback raises rather than returning false — is the pending send guaranteed not to go out, and is the guard removed cleanly? If the contract is "return false to deny, do not raise", this should match its three siblings. I could not check this myself because the API is not in any Lich build yet.

- **81 of 779 Bigshot examples (10.4%) are environment-gated, and they are exactly the native-integration ones.** That is disclosed and unavoidable while #1575 is unmerged, so it is not a criticism. The specific thing worth flagging: `spec/scripts/bigshot/quick_tether_spec.rb:219`, "leaves the ordinary helper cast path without a new target selection", is a **legacy-path** regression test and it is behind the same `LICH_EXECUTION_GUARD_ROOT` skip as the native ones. `cmd_tether` was restructured for this PR (`scripts/bigshot.lic:10346`) — the `loop do … return … end` body became a `proc` that is `.call`ed — so the legacy path did change shape, and nothing in CI exercises it. I verified the refactor is semantically identical by executing both forms (`proc` non-local return unwinds the enclosing method on the `return` path and falls through on the `break` path, and `unless a and b` ≡ `unless a && b` under modifier precedence), so I am not claiming a bug. **Settles it:** can that one example be split so its legacy half runs unconditionally? Maintainer steer on this review: the call is the author's, but more spec coverage is preferred where it can be had.

---

## What this PR gets right

Worth naming, because these are the load-bearing decisions to protect during revision:

- **The global-reset ordering hazard is handled, not stumbled into.** `Bigshot.new` is constructed *before* `reset.call` (`scripts/bigshot.lic:2336` vs `:2350`), and `Bigshot#initialize` writes `$bigshot_debug = true` at `:6940`, which `bigshot_initialize_globals` then clobbers. Line `:2351` explicitly recomputes `$bigshot_debug` from the engine's own debug ivars immediately after the reset. That is a subtle trap, spotted and closed.
- **The legacy engine path really is untouched.** Every in-method change is gated on `@quick_native_scope`/`@quick_profile` and returns early, and the two structural refactors (`cmd_tether`, `cmd_incant`) preserve semantics — I checked both empirically rather than by eye. `cmd_incant`'s new `ensure` even fixes a latent leak where `Spell.after_stance` was not restored if `cast_spell` raised.
- **Concurrency exclusion is enforced at the right layer.** A second `;bigshot quick clear` typed by the player is absorbed by the upstream control hook (`scripts/bigshot.lic:1861`) before Lich ever starts a script; one launched programmatically is refused by `owner_available?` (`:2180`). Both fail closed.
- **Negative UIDs are handled consistently with the fix that just landed underneath this branch.** `BSAreaRooms.for_quick` uses `/\Au-?\d+\z/` (`:4180`) and `QuickRetreat130.destinations` uses `/\A(?:[1-9]\d*|u-?[1-9]\d*)\z/i` (`:2535`), so Reim rooms work.
- **The GUI degrades instead of breaking.** `build_encounter_tab` rescues `ArgumentError`, sets `@encounter_settings = nil`, and appends an explanatory tab (`:5786-5791`), with `on_close_clicked` guarding on that nil. Since this tab is built for *every* user who opens Setup, that fallback matters. The `<object class="GtkNotebook">` string it patches does occur literally and first in `Setup.ui` (`:4948`), so the `sub` lands.
- **The developer documentation is in the right place.** `docs/` is the repo's home for developer documentation of a script, with GSWiki covering end-user usage — and all five new files are developer-facing by that split (`bigshot-quick-combat.md` opens "Development documentation:"; the lab contract, plan, smoke test and refuge notes are all implementation-facing). It has no build impact either way: `netlify.toml` publishes `dist` built by `jinxp -i scripts`, so `docs/` is not deployed. Noting this explicitly so it does not get re-raised — this review initially questioned the directory and was wrong to.
- **The specs exercise real production source.** They extract class bodies from `scripts/bigshot.lic` and `module_eval` them, with `gsub("\r\n", "\n")` for the file's CRLF endings, so shape changes fail loudly rather than silently drifting from a copy.

---

## Coverage notes

**Executed locally, at `0a9c971`:**

- `ruby -c scripts/bigshot.lic` → **Syntax OK**.
- `bundle exec rubocop` over all 56 changed Ruby files → **no offenses**.
- `bundle exec rspec spec/scripts/bigshot` → **779 examples, 0 failures, 81 pending**, matching the PR's claim exactly.
- `bundle exec rspec` (full suite) → **22,651 examples, 0 failures, 81 pending**, but only after running `bin/migrate` first to generate `dist/gameobj-data.xml`. Without it, 19,537 `spec/gameobj-data` examples fail; that is a local-environment artifact, not this PR. (The PR reports 22,685 examples; the 34-example gap is consistent with its run including the companion eLoot/go2 stack.)
- Cross-repo API verification against a local `elanthia-online/lich-5` checkout at `6303974c` (`LICH_VERSION = '5.20.1'`), plus release and PR state via `gh`.
- Behavioural comparison of the CLI gate by executing this commit's `QuickRequest.parse` against Lich's real argument tokenizer.
- `debug_msg` census by category across the whole file, and by enclosing class/module for every new `quick_*` method, to separate "could call `debug_msg` and doesn't" from "structurally cannot".
- YARD coverage measured on both revisions with the same script: every `def`, checking the contiguous comment block above it for `@param`/`@return`/`@raise`/`@yield`/`@option`. Base `273fa778` 299/299; head `0a9c971` 307/496.
- Per-tick cost of the observation-ring copy measured with a standalone benchmark of `EncounterPolicy.immutable_copy` over a full 100-entry status snapshot (~0.45 ms, ~1,541 allocations).

**Not verified, and why:**

- **Everything behind `LICH_EXECUTION_GUARD_ROOT` (81 examples).** The APIs do not exist in any Lich build, so no amount of local work covers the native execution-guard, attack-provenance, go2-child, walking-retreat, or real-GTK paths. That is the bulk of what makes this feature safe, and it is currently proven only by the author's live acceptance run.
- **Live game behaviour.** Refuge outings, seek movement, retreat, and the eLoot room-scoped loot path all move a character and send commands; nothing here substitutes for the player-authorized runs recorded in `docs/bigshot-quick-smoke-test.md`. That document's own stated limits (one route, no loot, one profile) remain the honest bound.
- **`spec/support/*` harness fidelity.** I read the specs' loading strategy and spot-checked several, but did not audit all 50 new spec files (11,068 added lines) for stubs that could mask a production regression.
- **eLoot #2457 and go2 #2458.** Reviewed only as they are consumed from Bigshot; their own diffs were out of scope.

**Prior automated review:** CodeRabbit reviewed this PR on GitHub and raised 7 comments; the author addressed the actionable ones in `fba781f` (guard-wrapper argument forwarding, the frozen-fixture provenance test, the `quick_safety_spec` extraction-order bug, markdown fence languages, and the ICT/UTC date qualification), and moved the eLoot empty-corpse-list finding to #2457. I re-checked those fixes at head and they are correctly applied. A local `coderabbit review --agent` pass was also run against the correct merge base (`273fa778`) and returned 4 findings. One became finding #12 above and one became finding #6. The other two I investigated and **dismissed** — recording them here because they will likely resurface on GitHub:

- *"`enqueue`'s rescue re-enters `@mutex.synchronize` (`scripts/bigshot.lic:1483`)."* Not a defect. `Mutex#synchronize` releases the lock in its own `ensure` as the exception unwinds, so the method-level rescue re-acquires a free mutex. Verified by execution: the rescue completes and records `@fault` with no `ThreadError`. The same shape in `arm` and `result` is equally fine.
- *"`CharSettings['untargetable'].include?` needs a nil guard (`scripts/bigshot.lic:12483`)."* Not a crash. lich-5 patches `NilClass#method_missing` to return `nil` (`lib/common/class_exts/nilclass.rb:9`), so `nil.include?(name)` evaluates to `nil` — falsy — which is behaviourally identical to the empty array. Worth knowing that this *is* newly reachable, though: the PR skips `CharSettings['untargetable'] ||= []` on the quick path (`scripts/bigshot.lic:6917-6922`), so a fresh character or one after `;bigshot reset` will hit the nil during a Quick run. It is safe today only because of that monkeypatch, which is worth a comment if you want it to stay safe.

Note that an earlier local run against the branch *name* `fix/bigshot-looting-watch-pause-race` resolved to a stale local copy (`5ba676b3`, not an ancestor of the PR head) and pulled in ~20 unrelated files; its output was discarded. If you run the CLI locally, pass `--base-commit`.

therealatari:
@mrhoribu The requested Bigshot Quick review changes are now pushed in `7dd5728e`. This restores permissive legacy Quick tails, adds lifecycle/action/error diagnostics, fixes the confirmed execution and GTK validation issues, updates YARD/docs/changelog language, and adds plain-language hover help to every Quick preset and named-trial field/control. The isolated suite is 783/783 green, the native integration suite is 783/783 green (five real-GTK cases gated there), and the five Xvfb GTK cases pass. The PR remains intentionally stacked on #2433, so the current conflict is expected until that stack lands; it also remains explicitly blocked on release of the Lich APIs in #1575/#1576/#1577.

therealatari:
Closing this expansion in favor of the modular EO Hunter path.

This branch was valuable: it proved the bounded execution, provenance, room-scoped cleanup, supervised refuge, and controller contracts in live use. The reusable seams remain independently reviewable in elanthia-online/scripts#2457 and elanthia-online/lich-5#1575, #1576, and #1577.

EO Hunter now provides the smaller behavior/arbiter architecture, group hunting, controller integration, and subsecond target/state response this use case needs. The focused Hunter follow-ups retain the lifecycle fixes in Nisugi/hunter#50 and optional Recorder trial context in Nisugi/hunter#51 without carrying a second 20k-line combat subsystem inside Bigshot. Keeping this 62-file PR open would duplicate the implementation and impose unnecessary review and maintenance cost.

This closes only the proposed Quick expansion; it is not a judgment on existing Bigshot behavior or compatibility.

PR #2464, opened 2026-09-11
feat(effect-list): cooldowns for spells that lock out their targets

Nisugi:
## Summary

Five spells stop a character from receiving them again for a while, and nothing in the effect list said so. Companion to elanthia-online/lich-5#1597, which reads these and tracks the cooldowns per character; without that PR these attributes are inert, and without this one its code has no data to act on. Either can merge first.

Two shapes, so two attributes.

**`group-cooldown`** -- the lockout a group (EVOKE) casting puts on each target. Bravery (211) and Heroism (215) at 180s, Spell Shield (219) at 360s. The caster's own start message names the group, which is how Lich tells a group casting apart from a self-cast.

**`target-cooldown`** -- the lockout a spell puts on the one character it lands on. Wall of Force (140) at 270s, Celerity (506) at 240s. It belongs to that character whoever cast it, so it pairs with a **`target-start`** message: the third-person line naming them, with the name captured as `(?<noun>...)`.

```xml
<spell ... number='140' ... target-cooldown='270'>
   <message type='target-start'>A wall of force surrounds (?&lt;noun&gt;[A-Z][a-z]+)\.</message>
```

Lengths are from the spell pages on the wiki. Existing attributes and messages are untouched, so a Lich that does not know these names ignores them.

Deliberately left out: Rapid Fire (515), whose recovery starts when the effect ends and scales with the target's Elemental Mana Control rather than being a constant; Barkskin (605), whose cooldown starts when the spell absorbs a hit, an event with no fixed relationship to the casting; and Invisibility (916) and Resist Nature (620), whose group versions have no per-target lockout.

## Test plan

- [x] XML parses under Ox with the same options Lich uses (`mode: :generic, skip: :skip_none`)
- [x] All five entries read back with the expected attribute values and messages
- [x] The two `target-start` patterns match the real game lines and capture the name, and do not match the second-person lines, the group-casting line, or a pet's line
- [x] In game on a live install: Wall of Force cast on a character reads 268.98 seconds left a moment later, from 0.0 before

### DIFF
--- a/scripts/effect-list.xml
+++ b/scripts/effect-list.xml
@@ -129,12 +129,13 @@
    <spell availability='all' name='Searing Light' number='135' type='attack'>
       <cost type='mana'>7</cost>
    </spell>
-   <spell availability='all' name='Wall of Force' number='140' type='defense' target-cooldown='270'>
+   <spell availability='all' name='Wall of Force' number='140' type='defense'>
       <duration cast-type='self' span='refreshable'>1.5</duration>
       <duration cast-type='target' span='refreshable'>1</duration>
       <cost type='mana'>40</cost>
       <bonus type='bolt-ds'>100</bonus>
       <bonus type='physical-ds'>100</bonus>
+      <cooldown type='target'>270</cooldown>
       <message type='start'>A wall of force surrounds you\.</message>
       <message type='end'>The wall of force disappears from around you\.</message>
       <message type='target-start'>A wall of force surrounds (?&lt;noun&gt;[A-Z][a-z]+)\.</message>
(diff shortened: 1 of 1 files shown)


mrhoribu:
would it be better served to do:
```xml
   <spell availability='all' name='Wall of Force' number='140' type='defense'>
      <duration cast-type='self' span='refreshable'>1.5</duration>
      <duration cast-type='target' span='refreshable'>1</duration>
      <cost type='mana'>40</cost>
      <bonus type='bolt-ds'>100</bonus>
      <bonus type='physical-ds'>100</bonus>
      <cooldown type='target'>270</cooldown>
      <cooldown type='group'>270</cooldown>
      <message type='start'>A wall of force surrounds you\.</message>
      <message type='end'>The wall of force disappears from around you\.</message>
      <message type='target-start'>A wall of force surrounds (?&lt;noun&gt;[A-Z][a-z]+)\.</message>
   </spell>
```

Nisugi:
Agreed, that reads better and matches how `duration`, `cost` and `bonus` are already carried. Adopted in 1e0d03c6 — all five spells now use `<cooldown type='group'>` / `<cooldown type='target'>` children, placed after `bonus` and before `message` as in your example.

Parsing side updated to match in elanthia-online/lich-5#1597, which now reads them with a `locate('cooldown')` loop instead of attribute lookups.

Worth landing the lich-5 side first or close to this one. Each is harmless alone, but if this merges first, anyone pulling a fresh effect-list before the parsing change ships would see these cooldowns read as nil.

mrhoribu (approved):
would it be better served to do:
```xml
   <spell availability='all' name='Wall of Force' number='140' type='defense'>
      <duration cast-type='self' span='refreshable'>1.5</duration>
      <duration cast-type='target' span='refreshable'>1</duration>
      <cost type='mana'>40</cost>
      <bonus type='bolt-ds'>100</bonus>
      <bonus type='physical-ds'>100</bonus>
      <cooldown type='target'>270</cooldown>
      <cooldown type='group'>270</cooldown>
      <message type='start'>A wall of force surrounds you\.</message>
      <message type='end'>The wall of force disappears from around you\.</message>
      <message type='target-start'>A wall of force surrounds (?&lt;noun&gt;[A-Z][a-z]+)\.</message>
   </spell>
```

PR #2469, opened 2026-09-13
fix(eloot): v2.11.9 massive boulders skinned with the wrong tool type

mrhoribu:
## Summary
- \`Loot.skin\` routes stone-hided creatures (krynch, stone mastiff, krag dweller, cavern urchin) to the blunt-weapon skinner instead of the default blade, since a blade doesn't work on their hides.
- \`massive boulder\` is the same kind of stone-hided creature but was missing from both the blunt-match and the blunt-exclude regex, so it fell through to the regular (bladed) skinner.
- Added it to both regexes so it's routed to the blunt skinner like its fellow stone-hided creatures.

## Test plan
- [x] `bundle exec rubocop scripts/eloot.lic` — no offenses
- [x] `bundle exec rspec spec/scripts/eloot_spec.rb` — 134 examples, 0 failures

### DIFF
--- a/scripts/eloot.lic
+++ b/scripts/eloot.lic
@@ -15,9 +15,12 @@
               game: Gemstone
               tags: loot
           required: Lich >= 5.15.0
-           version: 2.11.8
+           version: 2.11.9
   Improvements:
   Major_change.feature_addition.bugfix
+  v2.11.9 (2026-09-12)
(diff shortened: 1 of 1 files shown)


mrhoribu:
Closing — this fix is being folded into #2467 instead.

PR #2473, opened 2026-09-14
fix(eloot): v2.11.9 Hinterwilds sell run no longer walks you to Icemule for the pawnshop

Nisugi:
## Summary

Reported: a character hunting out of Coldriver Village (the Hinterwilds) is now taken to Icemule Trace to sell, where eloot used to just skip the pawnshop because the town doesn't have one.

The Hinterwilds town (uid `7503205`) has exactly three shop tags across its 246 rooms: a gemshop, a furrier, and the locksmith pool. `Sell.go_sell` has always known this and skipped `pawnshop|collect|consignment|chronomage` there unless "Sell in FWI" is on — that's the skip being remembered, and it still works.

The problem is that `go_sell` is not the only way to reach a shop. Several paths find one with their own `find_nearest_by_tag(shop)` call, which searches the entire map graph with no town or distance scoping, and then call `ELoot.go2` directly. From the Hinterwilds, the nearest pawnshop is Icemule's room 2464.

## Affected paths

| Path | Added | How it's reached |
|---|---|---|
| `Sell.retry_wrong_shop_jewelry_at_pawnshop` | v2.11.7 (#2453) | every sell run |
| `Sell.recheck_refused_at_pawnshop` | v2.10.0 (#2365) | opt-in "Recheck gemshop-refused items" |
| `Loot.sell_box_contents` | v2.11.8 (#2455) | locksmith pool full-box recovery |
| `Sell.custom_type` | — | `;eloot type <x>` |
| `Sell.custom_list` | — | `;eloot sell <item>` |

The first is the one being reported. Its trigger fits the location: the Hinterwilds gemshop is the only gem buyer in town, so any jewelry-tagged item the jeweler answers *"that's not quite my field"* to gets queued and dragged to Icemule. Before v2.11.7 the `jewelry` + `Pawnshop` guard in `Sell.appraise` stopped those items cold and they were simply stowed.

`sell_box_contents` is worth calling out: the Hinterwilds **has** a locksmith pool, so that recovery runs exactly where the bug lives. A pool box holding a pawnshop-only item walks you out of the area mid-hunt.

## What changed

Guarding call sites one at a time is what caused this in the first place — v2.11.7 added a shop trip and forgot the check — so the guard is folded into the lookup rather than bolted onto each caller:

- `ELoot.shop_unavailable_in_town?` — `go_sell`'s inline town check, extracted.
- `ELoot.nearest_shop(shop)` — applies that guard, then `find_nearest_by_tag`. Returns `nil` for "don't go". Every path that travels to a shop now finds it through here, so a new sell path gets the town scoping by construction instead of having to remember a separate check.

The guard was deliberately **not** put inside `ELoot.go2`. That method is also called with bare room ids for banks, start rooms and locksmith pools, so a check there would silently skip legitimate travel while callers carried on as if they'd arrived.

Behaviour notes:

- The two pawnshop retries keep two distinct outcomes. No pawnshop in town clears the queue, since it can never succeed there. No *reachable* pawnshop leaves it queued for a later run, which is the pre-existing behaviour.
- "Sell in FWI" still overrides everywhere, so anyone who has opted into travelling to sell is unaffected.
- `Sell.collectibles` was not reachable unguarded (both callers already filter), but it had its own unguarded lookup, so it is now safe from any future caller. It also returns instead of calling `go2(nil)` when nothing is reachable.
- Incidental: `go_sell` is now nil-safe. The old inline check evaluated `Room[nil].uid` when no town could be found.

## Test plan

- [x] `ruby -c scripts/eloot.lic` — Syntax OK
- [x] `rubocop scripts/eloot.lic spec/scripts/eloot_spec.rb` — no offenses
- [x] `rspec spec/scripts/eloot_spec.rb` — 165 examples, 0 failures (134 on master)
- [x] Mutation-checked both ways: reverting `nearest_shop`'s guard fails the helper specs, and reverting `sell_box_contents` to its old lookup fails three of the new box examples. The new tests are not vacuous.
- [x] Verified against the shipped map data that the Hinterwilds carries only `gemshop`, `furrier` and `locksmith pool` tags, and that the nearest `pawnshop` is Icemule room 2464
- [x] **Verified in-game:** a live sell in the Hinterwilds stayed in the Hinterwilds. This is the reported bug, confirmed fixed on a real character.
- [x] **Verified in-game:** retrieving boxes from the locksmith pool in the Hinterwilds also stayed in the Hinterwilds. This exercises `Loot.sell_box_contents`, a separate code path from the sell run above, and it is the one the Hinterwilds actually triggers since the town has a locksmith pool.

New coverage: the `shop_unavailable_in_town?` predicate (each missing shop skipped, each present shop still routed, the FWI opt-in, no-town, argument handling), the `nearest_shop` lookup, four cases on the jewelry retry, and four on `sell_box_contents` confirming it no longer leaves town and still sells what the town can buy.


scripts/eloot.lic
```diff
@@ -3215,6 +3222,44 @@ module ELoot # Game utility type methods
...
   end
 
+  # The Hinterwilds town (Coldriver Village) has only a gemshop, a furrier and
+  # the locksmith pool -- no pawnshop, collectibles, consignment or chronomage.
+  # Any trip to one of those from here leaves the area entirely (the nearest
+  # pawnshop on the map is Icemule's), which is almost never what you want mid-
+  # hunt, so we skip those shops unless "Sell in FWI" is on and you've opted
+  # into travelling to sell. Shared by every shop-routing path so a new one
+  # can't quietly reintroduce the cross-continent walk.
+  HINTERWILDS_TOWN_UID = 7503205
```
mrhoribu: This will cause the dreaded "Constant already initialized" spam in the debug log. Can you guard this from re-declaration on subsequent runs? Could probably just do
```ruby
HINTERWILDS_TOWN_UID ||= 7503205
```

mrhoribu (changes requested):

mrhoribu (approved):

PR #2478, opened 2026-09-17
fix(bigshot): v5.16.2 keep attacking a creature made sympathetic mid-fight

mrhoribu:
## Summary

- `<crtrStatus>` is a full snapshot, not a delta: casting the Empath warding spell Sympathy 1120 on a hostile creature replaces its `hostile="1"` with `sympathetic="1"` in the very next tag, with no `hostile` flag at all - and nothing in that tag or the surrounding messaging reports the creature dead, fled, or otherwise removed from the fight. `bs_hostile_creatures` required `crtr_flag?(:hostile)` alone, so bigshot silently dropped the creature from its target roster and broke off the fight at the next room/`crtrStatus` refresh after Sympathy landed (the attached log's `sympathetic` tag arrives on a follow-up `look`, not instantaneously at cast resolution), on a creature the game never actually reported as gone.
- Fix: also treat `crtr_flag?(:sympathetic)` as a valid attack condition, but gated behind a new `ever_hostile?(id)` check rather than trusted on its own - a creature must have been asserted hostile by the game at least once before `sympathetic` alone can carry it. This rules out a creature that's sympathetic but was never actually a combat target (nothing in lich-5 documents whether `sympathetic` could ever appear on a creature the game never flagged hostile, so this stays conservative rather than assuming it can't happen).
- Deliberately not gated on `GameObj.targets` (the client's selected-target dropdown) - that's the same stale, event-driven source `bs_hostile_creatures` was migrated away from as ground truth in #2414, and it has no obligation to refresh on a hostile -> sympathetic transition either.
- `remember_hostile`/`ever_hostile?`/`hostile_seen_ids` add the minimal state this needs: a per-id "seen hostile" memory, pruned via `Creature[id]` (an O(1) registry lookup, done once per already-held id) so it rides the registry's own size cap/eviction instead of reimplementing one. Dead creatures are still excluded regardless of hostile/sympathetic/ever_hostile.

Bumped to v5.16.2 with a changelog entry.
- Widened blast radius, intentionally: because `bs_targets` sits downstream of `bs_hostile_creatures`, a sympathetic-but-once-hostile creature now also counts toward `should_flee?`'s flee count, `gameobj_npc_check`, the `@MSTRIKE_MOB` threshold checks, and the `uncover()` empty-room guards - not just "keep attacking." This is intentional, not an oversight the changelog's one-liner undersells - it's the same roster every one of those already reads from - but it rests on the creature actually still being a threat post-flip, which the log doesn't directly show (see the open question below). If that turns out false, the fix is narrowing `bs_targets`' downstream consumers, not this gate.
- **Confirmed via live play (author's own testing): `hostile` comes back once Sympathy expires** - the reclassification doesn't stick for the fight. This means the cold-start case (a creature that's already `sympathetic` the very first time bigshot ever sees it, so `ever_hostile?` never got primed) self-heals on its own once the effect wears off and `hostile` reasserts, rather than being a permanent miss. Documented in the code comment above `bs_hostile_creatures` and covered by a new spec exercising the full lifecycle (sympathetic-only & excluded -> hostile reasserts & included -> sympathetic again & still included).

- **Also on this branch (not part of the targeting fix, flagged for visibility):** a commit removing the `type: 'bigshot'` parameter from a `strip_xml` call on the rest-decision path (`scripts/bigshot.lic:2841`) - it was forcing `Lich::Common::Markup.strip_xml` into multiline buffering it doesn't need for a single-line match, per review this also means a `DownstreamHook.remove` on bigshot stop wasn't clearing the stale multiline buffer for a subsequent run. And a fix making `COUP_INCAP_STATUSES` use `||=` like bigshot's other two class constants (`PRONE`, `PRONE_STATUSES`) already did, avoiding a constant-redefinition warning on a re-`;bigshot` - changelog entry added to match; note its "edit here" comment now also says a live edit needs a full Lich restart to take effect, since `||=` means a re-`;bigshot` won't pick it up.
- **Not fixed here, flagged for a separate PR:** `COUP_INCAP_STATUSES` includes `'bound'`, but nothing in lich-5 ever sets a `bound` status (the closest is `'bind'` in `STATUS_DURATIONS`) - so that entry can never match and the Coup de Grace R×10% threshold never unlocks from it. Pre-existing, on a line only touched here for the `||=` fix; leaving it alone rather than bundling a third drive-by fix into a PR about targeting.

## Open questions (unresolved, noted rather than guessed at)

- **Does the creature keep attacking after the hostile -> sympathetic flip, and does bigshot's own next attack against it restore `hostile="1"` immediately?** The captured log below only covers the run-up to the cast, not what happens after - it does not actually show continued attacking post-flip. If a bigshot attack restores `hostile` right away, the gap `ever_hostile?` covers is a fraction of a second rather than a full room refresh, and the widened blast-radius bullet above matters even less than described. Settled by: one capture of a bigshot attack landing on a sympathetic creature, followed by the next `<crtrStatus>` for that id.
- **Can either of the other two rest-trigger messages on the `strip_xml`-touched line arrive inside a suppressed stream (`inv`/`bounty`/etc. push-pop)?** Without the `type:` key, a line inside such a block is now matched on its own rather than deleted via `SUPPRESSED_STREAMS`. Believed to be no (lich-5 treats the item-cap message as a plain main-stream line), which is why this is a question and not a finding. Settled by: one capture of a full over-capacity loot cycle showing where those lines land relative to a `<pushStream id="inv"/>`.

## Evidence
```xml
<!-- CLIENT --><c>nw<!-- ENDCLIENT -->
<dialogData id='combat'><dropDownBox id='dDBTarget' value="sea nymph" cmd="target %dDBTarget%" tooltip='Select Target' content_text="sea nymph,sea nymph" content_value="#18156644,#18156647" anchor_left='cmdTarget' anchor_right='cmdAttack' height='20' width='80'  top='93' left='0' align='n'/></dialogData>
<nav rm='67018'/><streamWindow id='main' title='Story' subtitle=" - Coastal Cliffs" location='center' target='drop'/><streamWindow id='room' title='Room' subtitle=" - Coastal Cliffs" location='center' target='drop' ifClosed='' resident='true'/><clearStream id='room'/><pushStream id='room'/><compDef id='room desc'>A large <a exist="-28491" noun="tree">tree</a> clings tenaciously to the edge of the cliff, its roots dug deep into the face of the rock.  Branches like skeletal fingers form an arch overhead, and only a sparse scattering of yellowish-green leaves attests to the fact that it still lives.</compDef><compDef id='room objs'>  You also see<crtrStatus exist="18156647" hostile="1" inferior="1"/><b> <pushBold/>a <a exist="18156647" noun="nymph">sea nymph</a><popBold/></b>.</compDef>
<compDef id='room players'></compDef>
<compDef id='room exits'>Obvious paths: <d>southeast</d>, <d>west</d></compDef>
<compDef id='sprite'></compDef><popStream id='room'/>
<roommeta weather="0" bonfire="0" inside="0" water="0" sanctuary="0" realm="0" climate="4" terrain="10"/>
<component id='room players'></component>
<resource picture="0"/><style id="roomName" />[Coastal Cliffs]
<style id=""/><style id="roomDesc"/>A large <a exist="-28491" noun="tree">tree</a> clings tenaciously to the edge of the cliff, its roots dug deep into the face of the rock.  Branches like skeletal fingers form an arch overhead, and only a sparse scattering of yellowish-green leaves attests to the fact that it still lives.  <style id=""/>You also see <crtrStatus exist="18156647" hostile="1" inferior="1"/><pushBold/><pushBold/>a <a exist="18156647" noun="nymph">sea nymph</a><popBold/><popBold/>.
Obvious paths: <a exist="-11126286" coord="2524,1864" noun="southeast">southeast</a>, <a exist="-11126286" coord="2524,1864" noun="west">west</a>
<compass><dir value="se"/><dir value="w"/></compass><prompt time="1789602561">&gt;</prompt>
<dialogData id='encum'><progressBar id='encumlevel' value='0' text='None' top='5' left='-5' align='n' width='160' height='15'/><label id='encumblurb' value='You are not encumbered enough to notice.' top='10' left='0' align='n' width='160' height='50' justify='0' anchor_top='encumlevel'/></dialogData><component id='room players'></component>
<pushBold/>A <a exist="18156647" noun="nymph">sea nymph</a><popBold/> swings a dagger at you!
With incredible finesse, you deflect the attack with your <a exist="18151420" noun="shield">reinforced shield</a>!
<prompt time="1789602562">&gt;</prompt>
<dialogData id='minivitals'><skin id='healthSkin' name='healthBar' controls='health' left='0%' top='0%' width='25%' height='100%'/><progressBar id='health' value='100' text='health 195/195' left='0%' customText='t' top='0%' width='25%' height='100%'/></dialogData><dialogData id='injuries'><skin id='healthSkin' name='healthBar2' controls='health2' align='n' top='160' width='140' left='0' height='15'/><progressBar id='health2' value='100' text='health 195/195' customText='t' align='n' top='160' width='140' left='0' height='15'/></dialogData><dialogData id='minivitals'><skin id='manaSkin' name='manaBar' controls='mana' left='25%' top='0%' width='25%' height='100%'/><progressBar id='mana' value='98' text='mana 555/563' left='25%' customText='t' top='0%' width='25%' height='100%'/></dialogData>You feel more refreshed.
<prompt time="1789602562">&gt;</prompt>
<dialogData id='Buffs' clear='t'></dialogData><dialogData id='Buffs'><progressBar id='115' value='99' text="Fasthr's Reward" left='22%' top='0' width='76%' height='15' time='04:07:58'/><label id='l115' value='4:07 ' top='0' left='0' justify='2' anchor_right='incant'/></dialogData>
<dialogData id='Buffs' clear='t'></dialogData><dialogData id='Buffs'><progressBar id='115' value='99' text="Fasthr's Reward" left='22%' top='0' width='76%' height='15' time='04:07:58'/><label id='l115' value='4:07 ' top='0' left='0' justify='2' anchor_right='incant'/></dialogData>
<dialogData id='minivitals'><skin id='manaSkin' name='manaBar' controls='mana' left='25%' top='0%' width='25%' height='100%'/><progressBar id='mana' value='96' text='mana 545/563' left='25%' customText='t' top='0%' width='25%' height='100%'/></dialogData><dialogData id='combat'><cmdButton id='cmdTarget' value='target' cmd='target random' tooltip='Select a Random Target' height='20' width='55' top='93' left='0' align='nw'/><cmdButton id='cmdAttack' value='attack' cmd='attack' height='20' width='50' left='0' top='93' tooltip='Attack Current Target' align='ne'/></dialogData>
<dialogData id='combat'><dropDownBox id='dDBTarget' value="sea nymph" cmd="target %dDBTarget%" tooltip='Select Target' content_text="sea nymph" content_value="#18156647" anchor_left='cmdTarget' anchor_right='cmdAttack' height='20' width='80'  top='93' left='0' align='n'/></dialogData>
<castTime value='1789602567'/>You focus your thoughts while chanting the mystical phrase for Sympathy...
<a exist="spell" noun="spell">Your spell is ready.</a>
You gesture at <pushBold/>a <a exist="18156647" noun="nymph">sea nymph</a><popBold/>.
Closing your eyes and expanding your mind outward, you attempt to instill a sense of sympathy into those within the area...
  CS: +530 - TD: +6 + CvA: +25 + d100: +74 == +623
  Warding failed!
<pushBold/>A <a exist="18156647" noun="nymph">sea nymph</a><popBold/>'s eyes begin to glow ale-brown.
You open your eyes, feeling less focused.
Cast Roundtime 3 Seconds.
<prompt time="1789602564">&gt;</prompt>
<dialogData id='Buffs' clear='t'></dialogData><dialogData id='Buffs'><progressBar id='115' value='99' text="Fasthr's Reward" left='22%' top='0' width='76%' height='15' time='04:07:48'/><label id='l115' value='4:07 ' top='0' left='0' justify='2' anchor_right=''/></dialogData>
<prompt time="1789602574">&gt;</prompt>
<!-- CLIENT --><c>look<!-- ENDCLIENT -->
<roommeta weather="0" bonfire="0" inside="0" water="0" sanctuary="0" realm="0" climate="4" terrain="10"/>
<resource picture="0"/><style id="roomName" />[Coastal Cliffs]
<style id=""/><style id="roomDesc"/>A large <a exist="-28491" noun="tree">tree</a> clings tenaciously to the edge of the cliff, its roots dug deep into the face of the rock.  Branches like skeletal fingers form an arch overhead, and only a sparse scattering of yellowish-green leaves attests to the fact that it still lives.  <style id=""/>You also see <crtrStatus exist="18156647" sympathetic="1" inferior="1"/><pushBold/><pushBold/>a <a exist="18156647" noun="nymph">sea nymph</a><popBold/><popBold/> that appears sympathetic.
Obvious paths: <a exist="-11126286" coord="2524,1864" noun="southeast">southeast</a>, <a exist="-11126286" coord="2524,1864" noun="west">west</a>
<compass><dir value="se"/><dir value="w"/></compass><prompt time="1789602576">&gt;</prompt>
<!-- CLIENT --><c>look<!-- ENDCLIENT -->
<roommeta weather="0" bonfire="0" inside="0" water="0" sanctuary="0" realm="0" climate="4" terrain="10"/>
<resource picture="0"/><style id="roomName" />[Coastal Cliffs]
<style id=""/><style id="roomDesc"/>A large <a exist="-28491" noun="tree">tree</a> clings tenaciously to the edge of the cliff, its roots dug deep into the face of the rock.  Branches like skeletal fingers form an arch overhead, and only a sparse scattering of yellowish-green leaves attests to the fact that it still lives.  <style id=""/>You also see <crtrStatus exist="18156647" sympathetic="1" inferior="1"/><pushBold/><pushBold/>a <a exist="18156647" noun="nymph">sea nymph</a><popBold/><popBold/> that appears sympathetic.
Obvious paths: <a exist="-11126286" coord="2524,1864" noun="southeast">southeast</a>, <a exist="-11126286" coord="2524,1864" noun="west">west</a>
<compass><dir value="se"/><dir value="w"/></compass><prompt time="1789602581">&gt;</prompt>
```

## Test plan

- [x] `bundle exec rspec spec/scripts/bigshot_spec.rb` - 89/89 passing, including: sympathetic-after-hostile is kept, sympathetic-but-never-hostile is dropped, sympathetic+dead is still dropped even if once hostile, `hostile_seen_ids` forgets an id once Creature's registry no longer knows it, `ever_hostile?` survives a room-roster clear while the registry still holds the id (pins pruning against the registry, not the room roster - verified this test fails against a room-roster substitution before confirming it passes against the real code), and the full self-heal lifecycle (sympathetic-only & excluded -> hostile reasserts & included -> sympathetic again & still included).
- [x] `rubocop scripts/bigshot.lic spec/scripts/bigshot_spec.rb` - no offenses.


mrhoribu (commented):
# PR Review: fix(bigshot): v5.16.2 keep attacking a creature made sympathetic mid-fight (elanthia-online/scripts#2478)

**Reviewed:** 2 files, +155/−13 at `d0f69d9` on 2026-09-17. Based on: full clone of `elanthia-online/scripts` plus a full clone of `elanthia-online/lich-5` (including the `v5.21.0` tag, the version bigshot declares as its minimum).
**Revised:** 2026-09-17, after the author added the captured `<crtrStatus>` log to the PR body. Code unchanged at `d0f69d9`; one open question resolved, one narrowed.
**Verdict:** Approve with comments
**Summary:** The fix is correct, and with the log now attached it is correct on evidence rather than on inference — the capture shows `hostile="1"` on a sea nymph, a Sympathy cast, and the same creature id coming back `sympathetic="1"` with no `hostile` attribute at all. Nothing here blocks a merge. The one thing worth changing before it lands is the pruning loop in `hostile_seen_ids`, which does avoidable O(seen × registry) work on the hottest path in the script.

## Minor

### 1. `hostile_seen_ids` re-prunes with a linear scan on every single access — `scripts/bigshot.lic:8279`

```ruby
def hostile_seen_ids
  @hostile_seen_ids ||= {}
  live_ids = Creature.all.map { |c| c.id.to_i }
  @hostile_seen_ids.select! { |id, _| live_ids.include?(id) }
  @hostile_seen_ids
end
```

`live_ids` is an Array, so `include?` is a linear scan, and `select!` runs it once per remembered id — O(seen × registry) per call, plus a fresh `Array#map` over the whole registry to build `live_ids` in the first place. That happens even when `@hostile_seen_ids` is empty and there is nothing to prune.

The multiplier is what makes it worth fixing rather than shrugging at. `hostile_seen_ids` is reached once per hostile creature inside the `select` block at `scripts/bigshot.lic:8206-8207`, so a room with H hostiles prunes H times per `bs_hostile_creatures` call. `valid_target?` (`scripts/bigshot.lic:8657`) builds its `invalid_conditions` hash eagerly, so every call evaluates both `should_flee?` (which ends in `bs_targets.reject`, `scripts/bigshot.lic:8636`) and `still_targetable?` (`bs_hostile_creatures.any?`, `scripts/bigshot.lic:8317`) — two to three `bs_hostile_creatures` calls each. And `find_target` calls `valid_target?` once per candidate: `sort_npcs.each { |i| return i if valid_target?(i, just_entered) && priority(i) }` (`scripts/bigshot.lic:8793`). Five creatures in the room works out to roughly 75 prunes for one `find_target`.

The registry that gets scanned defaults to `max_size` 1000 with a 600-second `cleanup_max_age` (`lich-5/lib/common/creature/creature_base.rb:400,419`), and `Creature.all` returns `instances.values` — the whole registry, not just the room (`lich-5/lib/common/creature/creature_base.rb:285,509`). At a realistic mid-session registry of ~100 that is order 10⁵–10⁶ comparisons per `find_target`; at the 1000 cap it is two orders of magnitude worse. Not fatal, but it is pure waste in a loop that runs continuously during combat.

A Hash lookup fixes it in one line and keeps the same semantics:

```ruby
live = Creature.all.each_with_object({}) { |c, h| h[c.id.to_i] = true }
@hostile_seen_ids.select! { |id, _| live.key?(id) }
```

Better still, since the bound the pruning is defending against is slow-moving, prune on the same kind of wall-clock throttle lich-5 itself uses for `housekeep` (`lich-5/lib/common/creature/creature_base.rb:326,336`) rather than on every read. `@hostile_seen_ids` holds nothing but `Integer => true`, so even a few thousand unpruned entries is tens of KB.

### 2. The specs can't tell `Creature.all` from `Creature.in_room`, and that distinction is load-bearing — `spec/scripts/bigshot_spec.rb:723`

```ruby
def room(*fake_creatures)
  Creature.room_targets = fake_creatures
  Creature.registry = fake_creatures.each_with_object({}) { |c, h| h[c.id] = c }
end
```

The new stub at `spec/scripts/bigshot_spec.rb:688` carries a comment saying "real `Creature.all` is the full registry (room or not), not just the current room roster" — but `room` is the only thing that populates either, and it sets them to the same list. So in every test the registry *is* the room roster, and the difference the comment is documenting is never exercised.

Concretely: change `Creature.all` to `Creature.in_room` in `hostile_seen_ids` and all 87 examples still pass, including the new `forgets a creature id once Creature no longer knows about it` (`spec/scripts/bigshot_spec.rb:1193`), whose `bs.room` at line 1204 clears both at once. That substitution would be a real regression. `clear_room` empties the roster wholesale and the parser then re-marks creatures one at a time (`lich-5/lib/common/creature/creature_base.rb:253-266`), so a prune against `in_room` during that window would forget any creature not yet re-marked — and the next tag for it could be the `sympathetic` one, at which point it drops out of the target list for good.

The fix is small: give the harness a way to seat an id in `Creature.registry` without putting it in `room_targets`, then assert that `ever_hostile?` survives a roster clear while the registry still holds the id. That is the behavior the eviction story in the PR body actually depends on.

## Nits

- `scripts/bigshot.lic:8259` — `remember_hostile` is documented `@return [void]` but returns `true` (the value of the Hash assignment). Harmless, since the only call site discards it, but the neighbouring YARD in this file is accurate enough that this stands out.
- `scripts/bigshot.lic:8275` — the doc reference `{ClassMethods#cleanup_old}` won't resolve for anyone reading this from the `scripts` repo; `ClassMethods` is `Lich::Common::CreatureBase::ClassMethods` and lives in another project entirely. The surrounding comments in this same block cite lich-5 by path (`lib/gemstone/creature.rb`, `lib/common/xmlparser.rb`) — worth matching that here.

## Open questions

*The first item here in the original review — whether the game really drops `hostile` on this transition — is now answered by the attached log and has moved to the section below.*

- **Does `hostile` ever come back?** The capture covers roughly five seconds and two `look`s, both showing `sympathetic="1" inferior="1"` with no `hostile`. That is enough to prove the flip but not enough to say whether the flag returns when Sympathy expires, is dispelled, or the creature re-engages. It matters for one narrow case: a creature whose *first* `<crtrStatus>` bigshot ever sees is already `sympathetic` — a group empath sympathizing it before your character walks into the room — never primes `ever_hostile?` and is ignored for the rest of its life. **Settles it:** if `hostile` returns on expiry, that case self-heals and the gate costs nothing; if it does not, it is a permanent miss. Either way this is not a regression — pre-PR, `crtr_flag?(:hostile)` alone dropped that creature too — so it is a sentence in the comment at `scripts/bigshot.lic:8249-8255`, not a code change.
- **The flee/mstrike side effects are real and undisclosed.** Because `bs_targets` is downstream of `bs_hostile_creatures`, sympathetic creatures now also count toward `should_flee?`'s `flee_count` (`scripts/bigshot.lic:8649`), `gameobj_npc_check` (`scripts/bigshot.lic:6841`, which is `bs_targets`-backed despite the name), the `@MSTRIKE_MOB` thresholds at `scripts/bigshot.lic:6186,6203`, and the `uncover()` guards at `scripts/bigshot.lic:9590-9609`. The log supports this being the right call — the nymph swings a dagger at you while still hostile, and nothing in the capture suggests Sympathy stops it — but it is broader than "keep attacking," and the changelog entry only describes the attacking half. Worth one line in the PR body confirming it is intended.

## What this PR gets right

- **The attached log proves the premise outright.** The ordered sequence is unambiguous: `<crtrStatus exist="18156647" hostile="1" inferior="1"/>` on room entry, the Sympathy cast with `Warding failed!`, then the same id back as `<crtrStatus exist="18156647" sympathetic="1" inferior="1"/>`. `inferior="1"` riding through both tags is the detail that clinches it — it rules out a partial or truncated tag and shows `hostile` was genuinely cleared, exactly as `sync_crtr_status` treats an omitted classification flag (`lich-5/lib/common/creature/creature_base.rb:664-668`). One thing the capture does refine: the `sympathetic` tag arrives on a room refresh (here, a manual `look`), not spontaneously at cast time, so the fight actually breaks off at the next refresh rather than "the instant Sympathy landed" as the summary puts it. Worth correcting in the body — it is the kind of timing detail the next person debugging this will want to be exact.
- **The log also happens to prove the `GameObj.targets` argument.** The `dropDownBox` emitted right before `<nav rm='67018'/>` still carries `content_value="#18156644,#18156647"`, two sea nymphs, while the room the character actually arrives in contains exactly one (`exist="18156647"` in `compDef id='room objs'`). That is the client dropdown carrying a dead id across a room change — precisely the staleness #2414 migrated away from. The third bullet of the PR body asserts this; the evidence block now demonstrates it.
- **The `ever_hostile?` gate is the right shape.** `sympathetic` on its own is not evidence of hostility, and the conservative reading ("the game asserted hostile at least once") is defensible without needing to know whether a never-hostile creature can carry the flag. The PR body says as much instead of pretending to know.
- **Declining to gate on `GameObj.targets` is correct, and the reasoning is where the next person will look for it** — the comment block at `scripts/bigshot.lic:8153-8200`, not just the PR body.
- **Pruning against `Creature.all` rather than rolling a size cap.** Riding lich-5's own registry eviction is the right call; my objection above is to how the prune is implemented, not to the choice. (See Minor 2 for why it needs a test that actually pins it.)
- **`COUP_INCAP_STATUSES ||=` is a complete fix, not a partial one.** The only other class-level constants in `Bigshot` are `PRONE` and `PRONE_STATUSES` (`scripts/bigshot.lic:2732,2738`), and both already use `||=`. This brings the last one into line and matches the pattern merged in #2481.
- **The changelog entry is the right register for this repo** — one line, user-visible symptom, no internals. That is the convention, and it is easy to get wrong in the other direction.

## Coverage notes

- Re-reviewed on 2026-09-17 after the author attached the `<crtrStatus>` capture. No code changed (`d0f69d9` in both passes); only the grounding did. The log is the author's own capture and I have taken it at face value — I have no way to verify it was not edited, but it is internally consistent (matching `exist` ids, a coherent prompt-timestamp sequence, `inferior="1"` persisting across both tags) and it matches what lich-5's parser would do with it.
- Reviewed against full clones of both repos. `Creature.all`, `CreatureInstance.all` and the `sympathetic` classification key were each verified to exist at the `v5.21.0` tag that `bigshot.lic`'s `required:` header names, so the new call introduces no floor bump.
- **No Ruby toolchain on this machine**, so neither `bundle exec rspec` nor `rubocop` was run locally. GitHub's checks on `d0f69d9` are green ("Run tests on Ruby", "Run Rubocop on Ruby"), and the performance claim in Minor 1 is reasoned from call-graph structure and lich-5's documented defaults rather than measured — treat the magnitudes as estimates, not benchmarks.
- The claim in Minor 2 that substituting `Creature.in_room` would leave the suite green was derived by reading the two harnesses, not by running them. The harness code is short enough that I'd be surprised, but it is worth thirty seconds to confirm before acting on it.
- I traced the blast radius of `bs_hostile_creatures` across all fourteen call sites in `bigshot.lic`; `should_flee?`, `valid_target?`, `still_targetable?`, `sort_npcs`, `find_target` and `gameobj_npc_check` were each read in full. I did not audit the priority-ranking or looting paths beyond confirming they do not gate on hostility.
- Threading: I checked every `Thread.new` in `bigshot.lic` (lines 680, 700, 6104, 6656, 6763, 6790, 9973). None of them touch `bs_hostile_creatures`, so the in-place `select!` on `@hostile_seen_ids` has a single writer and I am not raising a concurrency concern.
- No human review or CI failure existed on the PR at the time of writing; the only comment is CodeRabbit reporting that auto-review is disabled.

mrhoribu:
Addressed the review:

- **Minor 1** (`hostile_seen_ids` O(seen × registry) prune): switched the live-id lookup from an `Array#include?` scan to a `Hash#key?` lookup.
- **Minor 2** (spec couldn't distinguish `Creature.all` from `Creature.in_room`): added `Harness#clear_room` (empties the roster only, mirroring real `Creature.clear_room`) and a test that primes `ever_hostile?`, clears the room while the registry still holds the id, and asserts it survives. Confirmed this test fails against a `Creature.all` → `Creature.in_room` substitution before confirming it passes against the real code.
- **Nit 1**: `remember_hostile` now returns `nil` explicitly, matching its `@return [void]` doc.
- **Nit 2**: the unresolvable `{ClassMethods#cleanup_old}` YARD reference now cites the lich-5 file path instead, matching the surrounding comments' style.
- **Open question 1** (already-sympathetic-on-first-sighting never primes `ever_hostile?`): documented as a known, non-regressing gap in the comment above `bs_hostile_creatures`.
- **Open question 2** (flee/mstrike/uncover blast radius): added a line to the PR summary confirming this is intended, not undisclosed.
- Also corrected an imprecise timing claim ("the instant Sympathy landed" → "at the next room/crtrStatus refresh"), both in the code comment and the PR summary, per the evidence log.

`bundle exec rspec spec/scripts/bigshot_spec.rb` — 88/88 passing. `rubocop` — clean. Pushed as `e79cde38` (merged cleanly with the `COUP_INCAP_STATUSES ||=` commit already on the branch).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

mrhoribu (commented):
# PR Review: fix(bigshot): v5.16.2 keep attacking a creature made sympathetic mid-fight (elanthia-online/scripts#2478)

**Reviewed:** 2 files, +226/−14 at `977d650` on 2026-09-18. Based on: full local clones of `elanthia-online/scripts` and `elanthia-online/lich-5`, with the Ruby 4.0.5 toolchain available — `bundle exec rspec spec/scripts/bigshot_spec.rb` (89/89 green) and `bundle exec rubocop scripts/bigshot.lic spec/scripts/bigshot_spec.rb` (clean) were both run locally, as was a mutation test of the new pruning spec.
**Verdict:** Approve with comments
**Summary:** The flag-flip this fixes is real and the code is sound — two rounds of self-review have already closed the pruning and spec-fidelity items. What is still open is the *reason* written into the source: the comment at `scripts/bigshot.lic:8189` asserts the sympathetic creature "is still fully engaged - still attacking," and neither the attached log nor lich-5's own bestiary supports that. That matters because the widened blast radius (flee counting, mstrike thresholds) follows directly from it.

## Minor

### 1. The `strip_xml` change is a second, unrelated behavior change with no test and no mention in the PR body — `scripts/bigshot.lic:2841`

```ruby
elsif server_string.is_a?(String) && strip_xml(server_string.dup) =~ /^You are unable to hold the number of items |...
```

Dropping `type: 'bigshot'` switches `Lich::Common::Markup.strip_xml` from `strip_multiline` to `strip_simple` (`lich-5/lib/common/markup.rb:107`). That is not cosmetic. With the type key, an unbalanced `<pushStream>` line was stashed in the process global `$strip_xml_multiline['bigshot']` and prepended to every subsequent line until a `<popStream>` balanced it; once balanced, `SUPPRESSED_STREAMS` could then delete everything between an `inv`/`bounty`/etc. push and its pop — including a target message that happened to fall in that window. The global also outlives the hook: `DownstreamHook.remove` on bigshot stop does not clear it, so a stop mid-stream left a stale prefix waiting for the next run.

So I think the removal is right, and lich-5 agrees about the shape of these messages — `lib/gemstone/combat/defs/messages.rb:125` matches `/^You are unable to hold the number of items /` against a plain main-stream line with no buffering of its own. But this arrives in a PR about creature targeting, is described only in the changelog and nowhere in the PR summary, and has no test. It is a one-line change to a live `DownstreamHook` on the rest-decision path; it deserves either its own PR or a sentence in the body saying what user-visible symptom prompted it. (`.dup` is also now redundant — `strip_simple` only ever calls `gsub`, which does not mutate — but that was true before this change too.)

### 2. `COUP_INCAP_STATUSES ||=` has no changelog entry, and the comment above it now misleads — `scripts/bigshot.lic:8455-8459`

```ruby
# (prone/kneeling/sitting) are deliberately excluded. Edit here if live
# testing shows the game honors more (or fewer) states.
COUP_INCAP_STATUSES ||= %w[stunned immobilized webbed sleeping bound].freeze
```

Two small things on the same line. First, #2481 — same author, same class of fix, merged two days earlier — bumped `briefcombat` to v1.0.5 *and* added `- Fix constant redefinition on subsequent runs` to its changelog. This PR makes the identical change to bigshot and lists only the sympathetic fix and the `strip_xml` one. Matching that precedent is a one-line addition.

Second, `||=` means an edit to this array no longer takes effect on a re-`;bigshot` — the constant is already initialized in the reopened class and the new value is discarded until Lich itself restarts. The comment directly above invites exactly that edit ("Edit here if live testing shows..."), so it should now say the change needs a Lich restart. The same caveat applies to `PRONE` and `PRONE_STATUSES` at `scripts/bigshot.lic:2733,2739`, which were already `||=`, but neither of those carries an invitation to edit them live.

## Nits

- **`hostile_seen_ids` still rebuilds the whole registry on every access** — `scripts/bigshot.lic:8300-8304`. The `Array#include?` → `Hash#key?` change removed the multiplicative term, but `Creature.all.each_with_object({})` is still an O(registry) build per call, even when `@hostile_seen_ids` is empty. Since the method only needs to know whether the handful of ids it *holds* are still live, `Creature[id]` — already used ten lines up at `scripts/bigshot.lic:8256` and O(1) via `instances[id.to_i]` (`lich-5/lib/common/creature/creature_base.rb:263`) — does the same job in O(seen): `@hostile_seen_ids.select! { |id, _| Creature[id] }`. Worth correcting the magnitude from the earlier round, though: I measured it rather than reasoning about comparison counts, and 75 prunes (one `find_target` with five creatures in the room) costs **7.3 ms at the 1000-entry cap and 0.79 ms at a realistic ~100**, against 0.02 ms for the O(seen) form. Real waste, trivially removable, but nowhere near the scale the previous review's "10⁵–10⁶ comparisons" framing implied. It is a cleanup, not a performance problem.
- **Side effect inside a `select` predicate** — `scripts/bigshot.lic:8220-8221`. `remember_hostile(c) if c.crtr_flag?(:hostile)` mutates state from inside a filter block and then `c.crtr_flag?(:hostile)` is evaluated a second time on the next line. Hoisting it (`hostile = c.crtr_flag?(:hostile); remember_hostile(c) if hostile`) reads better and halves the flag lookups on the hot path.
- **The first spec harness models `Creature.all` as the room roster** — `spec/scripts/bigshot_spec.rb:133` returns `(@room_targets || []).dup`, which is precisely the conflation the second harness's stub at `spec/scripts/bigshot_spec.rb:688` documents as wrong ("real `Creature.all` is the full registry (room or not)"). Nothing in harness 1 depends on the distinction today, so no test is wrong — but the two stubs of the same API now disagree, and the next person to add a pruning test in harness 1 will get a false pass.
- **Changelog is dated `2026-09-16`** (`scripts/bigshot.lic:20`) on a PR opened the 17th and still open on the 18th. The repo convention is the date of the change; harmless either way.

## Pre-existing (not caused by this PR, but on a line it touches)

- `'bound'` in `COUP_INCAP_STATUSES` (`scripts/bigshot.lic:8459`) can never match. `has_status?` is exact lowercase string membership in `@status` (`lich-5/lib/common/creature/creature_base.rb:625-628`), and no writer in lich-5 ever adds `"bound"` — `CRTR_STATUS_FLAGS` has no such key and `STATUS_DURATIONS` has `'bind'`, not `'bound'` (`creature_base.rb:71-101`). So the Coup R×10% threshold never unlocks from a bind. Unrelated to this change; flagging it because the line is in the diff and the comment above invites maintenance.

## Open questions

- **Can any of the three rest-trigger messages arrive inside a suppressed stream?** Without the type key, a line that is the middle of an `inv`/`bounty` push/pop block is now stripped on its own and its text can match, where the buffered form would have deleted it via `SUPPRESSED_STREAMS`. I believe the answer is no (lich-5 treats `^You are unable to hold the number of items ` as a plain main-stream line), which is why this is a question and not a finding. **Settles it:** one captured `;e echo` of a full over-capacity loot cycle showing where those lines sit relative to any `<pushStream id="inv"/>`.
- **Does attacking a sympathetic creature break the charm, and does bigshot's first swing therefore restore `hostile="1"` immediately?** If so, the flee-count window in Major 1 is a fraction of a second rather than nineteen, and the whole item drops to a comment fix. **Settles it:** one capture of an attack landing on a sympathetic creature, followed by the next `<crtrStatus>` for that id.

## What this PR gets right

- **The `Creature.all` vs `Creature.in_room` spec now actually pins the distinction.** I verified this rather than taking it on trust: substituting `Creature.in_room` for `Creature.all` at `scripts/bigshot.lic:8302` fails exactly one example — `spec/scripts/bigshot_spec.rb:1240`, "survives a room-roster clear while the registry still holds the id" — and nothing else. That closes the previous round's Minor 2 properly. The `clear_room` harness helper at `spec/scripts/bigshot_spec.rb:733` is the right shape for it, mirroring `lich-5/lib/common/creature/creature_base.rb:253-266`.
- **The `||=` fix is complete.** `bigshot.lic` has exactly three class-level constants — `PRONE`, `PRONE_STATUSES`, `COUP_INCAP_STATUSES` — and after this change all three are guarded. Verified by sweeping the file for constant assignment, not by reading the diff.
- **The `ever_hostile?` gate is conservative in the right direction.** Trusting `sympathetic` only after the game has independently asserted `hostile` costs one cold-start window and buys immunity from a whole class of false-positive target. Given that creatures themselves cast 1120 (`ashen_patrician_vampire.rb`, `ilvari_sprite.rb`, `glistening_cerebralite.rb`, `silver_scaled_cold_wyrm.rb`, `savage_fork-tongued_wendigo.rb` all list it as a warding spell), a creature carrying `sympathetic` without ever having been hostile to *you* is not hypothetical, and the gate is what keeps it out.
- **Pruning against the registry rather than rolling a size cap** remains the right call; it rides `cleanup_old`'s existing eviction and shelter rules (`creature_base.rb:307-325`) instead of duplicating them.
- **The changelog entries are in the repo's register** — user-visible symptom, one line, no internals. Easy to get wrong in the other direction.

## Coverage notes

- Full local clones of both repos, Ruby 4.0.5 available. `bundle exec rspec spec/scripts/bigshot_spec.rb` → 89 examples, 0 failures. `bundle exec rubocop scripts/bigshot.lic spec/scripts/bigshot_spec.rb` → 2 files, no offenses. The `Creature.all` → `Creature.in_room` mutation was run and produced one failure at `spec/scripts/bigshot_spec.rb:1240`.
- The 7.3 ms / 0.79 ms figures in the first nit are from a standalone Ruby benchmark of the prune loop shape against 1000- and 100-element registries, not from instrumenting bigshot in a live session. Treat them as the right order of magnitude.
- `Creature.all`, `Creature.[]` and the `sympathetic` classification key were each confirmed present at the `v5.21.0` tag bigshot's `required:` header names, so no Lich floor bump is implied.
- Threading re-checked at this head: all seven `Thread.new` sites in `bigshot.lic` (681, 701, 6105, 6657, 6764, 6791, 9994) were read; none touch `bs_hostile_creatures`, `bs_targets`, or `@hostile_seen_ids`, so the in-place `select!` has a single writer and there is no concurrency finding.
- What I did **not** verify: anything about Sympathy's in-game effect beyond lich-5's bestiary record and the attached log. I have no GemStone account and did not consult the wiki. Major 1 is built on those two sources and is labelled accordingly.
- The `strip_xml` change was traced through `lich-5/lib/common/markup.rb` and `lib/games.rb:1149` (reads are newline-delimited via `@socket.gets`, per `games.rb:337`), but not exercised against a real stream.

mrhoribu:
Addressed this round:

- **Minor 1** (`strip_xml` scope creep): that commit was pushed directly to the branch, not something I wrote - merged it in (fast-forward, no conflicts) and added a PR body section disclosing what it does and why, per the review's ask. Left it as-is rather than splitting into its own PR.
- **Minor 2** (`COUP_INCAP_STATUSES ||=`): added the missing changelog entry (matching the #2481 precedent) and updated its "edit here" comment to note a live edit now needs a full Lich restart, not just a re-`;bigshot`.
- **Nit 1** (`hostile_seen_ids` still O(registry) per call): replaced the `Creature.all` rebuild with `Creature[id]` - an O(1) registry lookup already used elsewhere in this file - per already-held id, so it's O(seen) now. Re-verified the pruning-source mutation test still fails against a regression before confirming it passes against the real code.
- **Nit 2** (double `crtr_flag?(:hostile)` eval in the `select` predicate): hoisted it to a local.
- **Nit 3** (first spec harness conflates `Creature.all`/`in_room`, unlike the second): left the harnesses as-is (only the second actually needs the distinction) but added a comment on the first harness's `all` stub flagging the conflation, so a future pruning test added there doesn't get a silent false pass.
- **Summary's core point** - the "still fully engaged, still attacking" claim wasn't actually supported by the log (the attack in it happens *before* the cast, not after): reworded the code comment and PR summary to say only what's evidenced (not reported dead/fled/gone), and added an "Open questions" section to the PR body for the two genuinely unresolved items (does attacking restore `hostile` immediately; can the `strip_xml`-affected messages land inside a suppressed stream) rather than asserting either way.
- **Pre-existing `'bound'` vs `'bind'` finding**: left unfixed, noted in the PR body as a separate follow-up rather than a fourth drive-by change in this PR.

One more thing this surfaced: I'd asked the user a follow-up on the open "does hostile ever come back" question from the previous round, and got a confirmed answer - **`hostile` does return once Sympathy expires**, live-play confirmed. That resolves the cold-start gap from a permanent miss to a self-healing one; documented in the code comment and covered by a new spec exercising the full lifecycle.

`bundle exec rspec spec/scripts/bigshot_spec.rb` — 89/89 passing. `rubocop` — clean. Pushed as `c9cbfeca` (merged cleanly on top of the `strip_xml` commit already on the branch).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2479, opened 2026-09-17
feat(map.lic): v3.0.0 rewritten from GTK to Lich's WebUI

Nisugi:
`;map` 3.0.0: rewritten from GTK to Lich's WebUI. Companion to the lich-5 WebUI stack (elanthia-online/lich-5 #1648 through #1655); it needs Lich with the WebUI at contract 2.19 and tells an older Lich so instead of failing.

## What changed

- The window is a browser app window served by Lich. The sheet, the room circle, the tag and location X markers and the note pins are drawn by the browser from data (composite layers), so they stay crisp at any zoom. Nothing is rasterised with Cairo any more.
- The options menu (Follow, Keep Centered, Expanded Canvas, Keep window on top, Dark Mode, Dynamic Indicator Size, Hide Scrollbars, Opacity, Scale, Maps, Tags, Locations, Find room, Add/Edit Note) is the contract's popup menu. Find room and the note editor are panels under the map instead of separate windows.
- Ctrl+wheel zooms around the pointer. Drag pans. Click, shift-click, ctrl-click and fix mode behave as before.
- Keep-on-top and opacity are applied to the real OS window. Borderless is gone: a browser app window cannot be made frameless, so the option is not offered.
- All the GC anchoring, `gtk_safe` wrapping and platform-specific window hints went with GTK. The script is about half its former length.

## What did not change

Every setting keeps its key (`keep_above`, `keep_centered`, `expanded_canvas`, `follow_mode`, `global_scale`, `global_scale_enabled`, `map_scale`, `dark_mode`, `opacity`, `hide_scrollbars`, `dynamic_indicator_size`), the per-character window size and position, `map_notes.json`, tags, locations, the links between sheets, `;map <room>`, `;map fix`, `;map trouble`, `;map reset` and `;map help`.

## How it was tested

- A harness that loads the script against a stubbed script scope and renders through Lich's real page builder and validator: sizing, layers, menus, click routing, notes, find, zoom, geometry persistence, settings, dark mode, centring through a real viewer attach.
- In game on Windows: opens at the saved size, follows the room, centres, zooms, keep-on-top and opacity applied.


mrhoribu (changes requested):
# PR Review: feat(map.lic): v3.0.0 rewritten from GTK to Lich's WebUI (elanthia-online/scripts#2479)

**Reviewed:** 1 file, +934/−2581 at `2c4c2b7` on 2026-09-17. Based on: full file at head, the previous `map.lic` at `master`, and the unmerged lich-5 WebUI stack checked out at `origin/webui/default-flip` (`28cd856f`, the head of lich-5 #1655, which contains #1648–#1655). Findings were verified by running `ElanthiaMap::Window#render` against lich-5's real `TreeBuilder`/`Validator`, and by counting the live GS4 mapdb (`elanthia-online/mapdb`, 21,497 rooms with an image across 276 sheets).
**Verdict:** Request changes
**Summary:** A careful, well-structured rewrite — the locking discipline is right, the tree validates, and the coordinate maths round-trips correctly against the client. But two of the options menu's own entries can produce a composite that exceeds the contract's 512-layer bound and kill the script, and the per-character window *position* that the PR says is preserved is never actually applied to the window.

## Blockers

### 1. Tags / Locations markers blow the contract's 512-layer bound and kill the script — `scripts/map.lic:659`

```ruby
marked_rooms(@active_tag) { |room| room.tags&.include?(@active_tag) }.each do |coords|
  result.concat(cross(coords, ox, oy, BLUE, s))
end
marked_rooms(@active_location) { |room| room.location.to_s == @active_location }.each do |coords|
  result.concat(cross(coords, ox, oy, GREEN, s))
end
```

`marked_rooms` (`map.lic:680`) is uncapped, and `cross` emits two `line` layers per room. With the sheet image and the room marker that is `2N + 2` layers for `N` marked rooms, while `Contract::COMPOSITE_PROPERTIES[:layers]` is `array(COMPOSITE_LAYER, max: BOUNDS[:collection])` — 512 (`lich-5/lib/webui/contract.rb:618`). So any selection covering **256 or more rooms on one sheet** raises `SchemaViolationError` out of `builder.composite`.

I drove the real `Window#render` through lich-5's real `Page`/`TreeBuilder`/`Validator` with a stubbed script scope. 255 marked rooms → 512 layers, renders. 256 → 514 layers → `Lich::WebUI::SchemaViolationError: must contain at most 512 items (cid=page:map/scroll:view/composite:map field=layers)`.

These selections are not hypothetical — they are entries this script's own menus build. Counting the current GS4 mapdb, seven `(sheet, tag-or-location)` groups are already over the line:

| Rooms | Entry | Sheet |
|---|---|---|
| 355 | Location: Old Ta'Faendryl | `en-old_ta_faendryl-1716923763.png` |
| 329 | Tag: ironfern root | `wl-gates-1264234799.png` |
| 308 | Location: the Citadel | `rr-citadel-1717856937.png` |
| 290 | Tag: ironfern root | `wl-mineroad-1264234799.png` |
| 272 | Tag: withered black mushroom | `wl-gates-1264234799.png` |
| 259 | Location: Ta'Illistim | `en-ta_illistim-new.jpg` |
| 258 | Location: Darkstone Castle | `wl-darkstone-1717380773.png` |

Notes on the sheet add two more layers each, so the real threshold is lower than 256 for anyone using notes.

What the player sees is worse than a one-off error, because it fails twice on different paths. `on_tag`/`on_location` (`map.lic:970`) deliberately don't call `refresh` — they lean on the runtime's post-`menu_item change` refresh, which runs in `Runtime#refresh_loop`; that rescues `StandardError`, logs, and gives up (`lich-5/lib/webui/runtime.rb:638`), so the window simply stops updating with no message. Then the next room change reaches `show_room` → `refresh` → `Lich::WebUI.refresh`, which has no rescue, so the exception unwinds out of the follow loop into `ElanthiaMap.run`'s handler and the script prints `[map: Fatal error: Lich::WebUI::SchemaViolationError: ...]` and dies. The only saving grace is that `@active_tag`/`@active_location` are not persisted by `save_settings`, so restarting `;map` recovers.

The menus themselves are already capped against this bound (`values.first(COLLECTION - 2)` at `map.lic:872`, `maps.first(COLLECTION - 2)` at `map.lic:815`), which is fairly good evidence the layer list was just missed rather than considered and accepted. A budget in `layers` — cap the combined marker set at `512 - (image + marker + note layers)` and `respond` once that markers were truncated — would be enough, though drawing each marked room as a single `rect` instead of two crossed `line`s doubles the headroom for free.

## Major

### 2. The saved window position is never restored — `scripts/map.lic:315`

```ruby
@browser = Lich::WebUI.open(page: @page, presentation: -> { window.window_presentation },
                            title: title_for(nil),
                            geometry: { width: @char_settings[:window_width].to_i, height: @char_settings[:window_height].to_i })
```

The PR body says "the per-character window size and position" is unchanged, and the comment at `map.lic:617` says "a reopened window comes back where the viewer left it". `window_position` is read at `map.lic:377`, updated from the browser at `map.lic:1073`, written back at `map.lic:389` and rendered into the geometry facility at `map.lic:622` — but nothing ever applies it:

- `BrowserLauncher.geometry_arguments` only emits `--window-position` when the geometry hash carries `position:` (`lich-5/lib/webui/browser_launcher.rb:119`). The call above passes only `width`/`height`, and `PresentedWindow.open` forwards the hash unchanged (`lich-5/lib/webui/presented_window.rb:31`).
- On the client, `fitWindow` reads only `geometry.width`/`geometry.height` (`lich-5/lib/webui/assets/app.js:1644`), `applyFacilities` reads only width/height (`app.js:1688`), and `moveTo` appears nowhere in `app.js`. The facility's `x`/`y` are inert.

So the window comes back at the right size and wherever Chrome feels like putting it. The fix looks like one line — add `position: @char_settings[:window_position].first(2).map(&:to_i)` to the `geometry:` hash, since the launcher already supports it — plus either dropping `x`/`y` from the facility or saying in a comment that they're carried for a client that may honour them later.

### 3. Keep-on-top and Opacity are now Windows-only, and fail silently elsewhere — `scripts/map.lic:325`

`window_presentation` feeds `PresentedWindow`, whose `apply` is a no-op until a window handle has been found (`lich-5/lib/webui/presented_window.rb:63`), and discovery only runs when `WindowPresentation.available?` — which is `false` unless `Lich::Common::Frontend.native_windows_runtime?` (`lich-5/lib/webui/window_presentation.rb:68`). The old version used `Gtk::Window#keep_above` and `#opacity` (`map.lic@master:642`, `:1125`), which worked wherever GTK ran.

On Linux and macOS, "Keep window on top" and the whole Opacity submenu now toggle, persist, re-render with the new check state, and do nothing at all. The `presentation` facility is also sent to the client, but the runtime records it as a degradation rather than honouring it for a real OS window. The changelog calls out Borderless as removed; this deserves the same treatment — either hide the two items when `Lich::WebUI::WindowPresentation.available?` is false, or say plainly in the changelog and `;map help` that they are Windows-only. "How it was tested" lists Windows only, which is consistent with this having gone unnoticed.

### 4. Every setting written from a menu handler is silently discarded — `scripts/map.lic:930`

`on_setting_toggle`, `on_follow_toggle`, `on_opacity`, `on_scale`, `on_surface_zoom` and `on_window_geometry` all call `save_settings`, which writes through `Settings[]=` / `CharSettings[]=`. Those run on the WebUI dispatcher thread, and `Settings` resolves its namespace from `Script.current`:

- `CharSettings.[]=` → `Settings.set_script_settings(active_scope, name, value)` with no `script_name` (`lich-5/lib/common/settings/charsettings.rb:18`), which falls through to `script_name || Script.current.name` (`lich-5/lib/common/settings.rb:365`).
- `Script.current` resolves by thread: `__running_snapshot.find { |c| c.has_thread?(Thread.current) }` over `@thread_group`, `@cleanup_thread_group` and borrowed workers (`lich-5/lib/common/script.rb:1077`, `:2876`, `:2557`).
- The dispatcher's worker thread is created by `Dispatcher#owner_state`'s `Thread.new` from a server connection thread (`lich-5/lib/webui/dispatcher.rb`), and nothing in `lib/webui/` binds it to a script — the only thread variable it sets is `THREAD_CONTEXT_KEY`. Because this script deliberately starts the service from `ThreadGroup::Default` (`map.lic:306`, which is the right call for the listener), that whole chain lives outside the script's group.

So `Script.current` is `nil`, `nil.name` is `nil` under lich-5's `NilClass#method_missing` patch, and `save_to_database` takes its `script_name.nil? || script_name.empty?` branch: *"Aborting save. ... Data will NOT be persisted"* (`lich-5/lib/common/settings.rb:367`). That log line is itself suppressed, since `@@log_level` defaults to `LOG_LEVEL_NONE`. Worse, `current_script_settings` first caches and mutates a root under the key `"::<scope>"` — a namespace belonging to no script.

In practice nothing is lost today, because `run` and the `before_dying` block both call `save_settings` from the script's own thread on the way out, and that write does land. But the design here is save-on-change, and it isn't happening: any exit that skips `before_dying` loses the session, and `on_window_geometry` fires a dropped write roughly twice a second while a window is being dragged. The `Settings` API has the seam you need — `Settings.root_proxy_for(scope, script_name: ...)` and the `script_name:` keyword on `set_script_settings` both accept an explicit name — so capturing `@script.name` in `initialize` and threading it through, or simply marking `@dirty` in handlers and doing the write from the follow loop on the script thread, would close it.

## Minor

- **`@pending_zoom_scroll` is never initialized** — `map.lic:1130` is its first assignment and `map.lic:1134` reads it via `.tap`, which is an uninitialized-ivar read on the path where `on_surface_zoom` returns early because `@current_map` is nil. Harmless (`nil` short-circuits the `if scroll`), but every other ivar in the class is declared in `initialize`; this one should be too, if only so the class reads consistently.
- **`register_sheets`'s rescue leaves two ivars undefined** — `map.lic:449`: if the first `register_files` raises, `@dark_dir` and `@dark_route` are never assigned, and `sheet_source` reads them. It works (both read as `nil`), but the rescue sets `@route = nil` explicitly and should set the other two for the same reason.

## Nits

- `window = self` is assigned twice in `show`, at `map.lic:295` and again at `map.lic:314`; the second is dead.
- The contract gate at `map.lic:62` requires `major == 2 && minor >= 19`. That matches `Contract::MAJOR_VERSION` today, so it is correct — but it means a future contract 3.x locks the script out with a message that says "2.19 or later", which will read as a bug when it happens.

## Open questions

- **Does the window grow by the title-bar height on every launch?** `windowGeometry()` reports `window.outerWidth`/`outerHeight` (`app.js:51`), which is what `on_window_geometry` stores and what `show`/the geometry facility hand back. But `fitWindow` treats the declared geometry as the *content* size and adds the chrome before resizing: `width = ceil(inner.width + chrome.width)` where `chrome = outerWidth - innerWidth` (`app.js:1651-1657`). Read literally that makes each `;map` come up one title bar taller than the last. It may never fire — Chrome blocks `window.resizeTo` on windows it did not open via script, and a `--app=` window launched from the command line is arguably one of those, which would match your "opens at the saved size" result. **Settles it:** start `;map` three times in a row without touching the window and note the height each time. If it grows, the fix is on the lich-5 side (report `innerWidth`/`innerHeight`, or stop adding chrome in `fitWindow`), not here.
- **Is one image layer per sheet enough for the dark-map set?** `sheet_source` (`map.lic:463`) falls back to the light sheet when the dark variant is missing, which is right — but it means a half-populated `maps-dark` directory silently mixes light and dark sheets in the same session. Was that the old behaviour too, or did the GTK version tint? (`resolve_map_path` at `map.lic@master:437` looks like the same fallback, so I suspect this is unchanged and fine.)

## What this PR gets right

- **The lock discipline is correct, and that is not easy here.** `render` holds `@lock` for the whole tree build, and every method the builder blocks reach back into — `layers`, `canvas_size`, `scale`, `tags_on_sheet`, `render_*_submenu` — is lock-free, so there is no re-entrant `Mutex` deadlock. Every handler that calls `refresh` releases `@lock` first (`on_surface_activate:1178`, `on_surface_zoom:1133`, `on_open_panel:980`). I looked for a deadlock specifically and could not construct one.
- **The zoom-anchoring maths is right.** `px = x + scroll_x` reconstructs the unscaled surface pixel exactly as the client intends (`app.js:1288-1303` divides both sides by the rendered scale before splitting them), and `px * new - x * old` is the correct new scroll offset to hold the point under the pointer. Easy to get subtly wrong; this isn't.
- **The whole component tree validates**, including the panels, the notes, the empty-map fallback and the no-notes case — I ran each through the real validator. Every prop used (`hidden` on `text_input`, `context_menu` on `text`, `variant`, `emphasis: 'subtle'`, `kind: 'separator'`, `compact`, `slot`) exists in contract 2.19, and the menus are already capped against `BOUNDS[:collection]`.
- **The comments explain the non-obvious choices rather than the obvious ones** — why the service starts from `ThreadGroup::Default`, why `scroll_position` is written per viewer and rendered once, why the builder blocks capture `window` and `builder` explicitly under `instance_exec`. Each of those was a real hazard and each is called out where the next reader will hit it.
- **The dark-map path is more robust than what it replaces**: `File.join(File.dirname(@map_dir), 'maps-dark')` handles Windows separators, where the old `map_path.sub('/maps/', '/maps-dark/')` would not have.

## Coverage notes

- No `gh` CLI in this environment; the PR was fetched over the GitHub API and raw content. The `scripts` repo was available as a local clone, so `master`'s `map.lic` and its history were read directly.
- **The entire lich-5 WebUI stack this depends on (#1648–#1655) is still open.** Everything above was verified against `origin/webui/default-flip` (`28cd856f`); if that stack changes before merge, the layer bound, the launcher's `--window-position` handling and `WindowPresentation.available?` are the three things to re-check. The two repos need to land together in either case.
- Findings 1 and 4 and the tree validation were confirmed by execution: the script's `ElanthiaMap` module was loaded against a stubbed script scope (`Room`, `Settings`, `CharSettings`, `XMLData`, `Char`, `respond`) and rendered through lich-5's real `Page`/`TreeBuilder`/`Validator` under Ruby 4.0.5. Finding 1's room counts come from the current `elanthia-online/mapdb` `gs/rooms` tree. Findings 2 and 3 are traced through code on both sides but not confirmed against a running Lich.
- Not verified: anything requiring a live game session or a real browser — the window-fitting question above, whether `PresentedWindow` finds the Chrome window by title in practice, actual click-to-`go2` routing, and `;map fix`. Rubocop was not run (this repo's bundle is not installed in this environment), so style-rule conformance is unchecked.

Nisugi:
Thanks for the review. Everything is addressed at `a89c8265` (plus `f9d3ecc0` for the rubocop failure CI caught).

**Blocker 1, layer bound.** `layers` now budgets against `COLLECTION`: the sheet, the notes and the room circle are always drawn, the markers get what is left at two layers each, and a selection that is cut short is announced once per (sheet, tag, location) so the follow loop does not repeat it. I kept the crossed lines rather than switching to a rect so the markers look as they did; the cap is what matters. Pinned by a harness case with a 401-room tag selection that renders within the bound.

**Major 2, position.** The saved position is now passed in the `geometry:` hash to `Lich::WebUI.open`, so the launcher emits `--window-position`. The facility's `x`/`y` are kept because the lich-5 client now honours them: Chrome ignores `--window-position` once it is already running, exactly as it ignores `--window-size`, so the client moves the window on the same first fit that sizes it (lich-5, `webui/review-2-fixes`).

**Major 3, keep-on-top and opacity.** The two menu items are only rendered when `Lich::WebUI::WindowPresentation.available?` is true, and the changelog says they are offered only where Lich can reach the OS window. Harness case renders the menu with the override both ways.

**Major 4, settings from the dispatcher thread.** Handlers no longer call `save_settings`; they set a dirty flag and the follow loop writes on the script's thread (it also wakes on a Follow toggle so the write is not left waiting for a room change). Harness asserts the handler leaves `CharSettings` untouched and the flush writes it.

**Minor and nits.** `@pending_zoom_scroll` declared in `initialize`; `register_sheets`' rescue clears all three route ivars; the dead `window = self` is gone; the gate message now says "within contract 2".

**Open question, window growth.** You were right, and it was not this script. The hidden geometry field reported `outerWidth`/`outerHeight`, `fitWindow` treated that as the content size and added the chrome, so each open grew by a title bar until `availHeight` clamped it. In-game: placed the window, reopened it, it grew once, then held. Fixed on the lich-5 side (the field reports the inner size now; position stays outer for `moveTo`), with a client-harness case. Dark-sheet fallback is unchanged from 2.x, as you suspected.

PR #2480, opened 2026-09-17
fix(eloot): stop retrying successful disk placement

therealatari:
## What changed

- treat the game's successful PUT/PLACE response as authoritative in Inventory.store_item
- stop retrying _drag while a disk's cached contents are still catching up
- add a regression built from the observed Skooshii disk loop

## Why

The game confirmed that a chest was placed in the disk, but the disk contents collection had not refreshed yet. Eloot retried the same _drag five times, repeatedly removing and replacing the chest, then falsely marked the disk full and moved the chest to a backpack.

## Verification

- regression was red before the fix: store_item returned false after five commands
- regression is green after the fix: store_item returns true after exactly one command
- spec/scripts/eloot_spec.rb: 166 examples, 0 failures
- ruby syntax check passes

A full unbundled repository run could not load three Jinx specs because the local Ruby environment lacks rack; the focused Eloot suite is complete and green.


mrhoribu:
By chance do you have a debug log of this behavior that could be analyzed. Something may need to be done for the GameObj disk object handling in Lich5 to make this not be an issue. Disks are always a bit of an oddity. But they should be getting reset and so something is likely off with it due to how it populates. Trying to match for all possible success patterns is a path to failure due to how much customization there is. Granted if only doing it for a single, known container type that has no special messaging possibilities (ie disks) then likely ok. But would love to try to fix the underlying issue rather than slap a bandaid on top.

PR #2482, opened 2026-09-17
feat(spellson.lic): v2.0.0 rewritten from GTK to Lich's WebUI

Nisugi:
`;spellson` 2.0.0: rewritten from GTK to Lich's WebUI. Companion to the lich-5 WebUI stack (elanthia-online/lich-5 #1648 through #1655); it needs Lich with the WebUI at contract 2.17 and tells an older Lich so instead of failing.

## What changed

- The window is a browser app window served by Lich. Every active spell is one row: the time left, a bar coloured by spell circle, the spell name over the bar. The bars are drawn by the browser from data (composite layers), so they stay crisp at any zoom.
- Keep-on-top is applied to the real OS window, as the GTK window was.
- The window size and position are reported by the browser and saved under the same Settings keys as before.
- No GTK, no `Gtk.queue`, no CSS providers. The row maths (stacking ceilings, duration captured on first sight, Indefinite past 599 minutes) is unchanged.

## What did not change

Settings keys `window_width`, `window_height`, `window_position`; the spell order; the per-circle colours.

## Two deliberate departures

- The death-decay timer (6666) is now salmon. The old palette keyed it as circle 6666, but the circle rule strips the last two digits, so it always fell through to the red fallback. It is keyed as 66 now.
- The GTK dark-theme palette is gone. The page has no theme signal, so the light palette ships at 0.85 alpha, which reads on both backgrounds.

## How it was tested

- Rendered through Lich's real page builder and validator with fake rows: facilities, layers and the hidden geometry field validate; the geometry event round-trips size and position into Settings.
- In game on Windows: opens at the saved size, keeps on top, one row per active spell coloured by circle, resizes without a horizontal scrollbar.


mrhoribu (changes requested):
# PR Review: feat(spellson.lic): v2.0.0 rewritten from GTK to Lich's WebUI (elanthia-online/scripts#2482)

**Reviewed:** 1 file, +301/-110 at `b7c4e61d` on 2026-09-17 (re-review; first pass was at `46442072`).
Based on: full clone of `scripts`, plus the `lich-5` WebUI stack read from the `pr-1657` ref (contract 2.19, the
head of the #1648-#1655 series) — `lib/webui/**`, `lib/webui/assets/app.js`, `lib/common/webui_launcher*`,
`lib/common/script_scope/gtk/session.rb`, `lib/common/settings*`, `lib/common/script.rb`, `lib/common/spell.rb`.
Rubocop with the repo config: clean. `ruby -c`: clean. No runtime/browser verification.
**Verdict:** Request changes
**Summary:** `b7c4e61d` correctly fixes both things it set out to fix. What remains is the plumbing around
opening the window: the service-start thread handoff is missing the gate lich-5's own copy of this code has, a
browser that fails to open is invisible, and the geometry the browser reports is not the geometry it is given
back.

## Fixed in `b7c4e61d`

- **Saved window position is now applied** (`scripts/spellson.lic:155-158`). `position:` reaches
  `BrowserLauncher.geometry_arguments`, which type-checks for exactly a two-Integer array and emits
  `--window-position` (`browser_launcher.rb:119-121`). Correct, and the comment naming the division of labour
  between the launcher and the client is accurate — nothing in `app.js` calls `moveTo`.
- **Settings are written from the script thread** (`spellson.lic:200-213`, `:319`). The dirty-flag handoff is
  right: `@settings_dirty.tap { @settings_dirty = false }` returns the pre-read value, so no flush is lost, and
  `save_settings` clearing the flag under the same lock means a concurrent handler update is written by the
  in-flight save rather than dropped. The commit message's diagnosis matches what I traced —
  `Script.current` resolves by thread-group membership (`script.rb:1075-1078`), the dispatcher thread is in
  `ThreadGroup::Default` by design, and `save_to_database` silently aborts on a nil script name
  (`settings.rb:365-370`) with logging off by default (`settings.rb:28`).

## Major

### 1. The service-start thread handoff races, and losing it takes the whole WebUI down — `scripts/spellson.lic:145-147`

```ruby
starter = Thread.new { Lich::WebUI.start }
ThreadGroup::Default.add(starter)
starter.join
```

`Thread.new` puts `starter` in spellson's thread group and starts it immediately; `ThreadGroup::Default.add`
runs afterwards. If `starter` reaches `Server#start` before the parent completes the move, the accept-loop
thread (`lib/webui/server.rb:93`, `@accept_thread = @thread_factory.call { accept_loop }`) inherits spellson's
group, and Lich kills that whole group when the script stops.

lich-5 has this same function, and it is written with a gate for precisely this reason
(`lib/common/script_scope/gtk/session.rb:276-297`):

```ruby
return service if service.server.running?
return service.start if Thread.current.group.equal?(ThreadGroup::Default)
gate = Queue.new
thread = Thread.new { gate.pop; result << [:ok, service.start] }
begin; ThreadGroup::Default.add(thread); rescue ThreadError; nil; end
gate << true
```

Its doc comment states the consequence: *"a server started from a session thread would lose its accept loop with
the first script to use it and could never rebind its port."* `Server#running?` is `@accept_thread&.alive?`
(`server.rb:562`), so after that every `launch_url` raises and every already-open WebUI window — the launcher,
any other script's page — is talking to a dead server. A later `start` does rebind, but on a fresh ephemeral
port, so those windows do not come back.

The narrowness cuts both ways. If the move *succeeds*, the dispatcher thread lands outside the script — which is
what `b7c4e61d` now correctly compensates for in the settings path. If it loses, the settings write would start
working and the WebUI would die instead. Copy `start_service`'s shape — gate, `running?` short-circuit,
`ThreadError` rescue — and both halves become deterministic.

Confidence: confirmed for the mechanism (Ruby assigns a new thread to its creator's group;
`__raw_worker_threads`, `script.rb:2556`, is literal group membership). Suspected for how often it actually
loses, since the parent's next operation is a C call that normally doesn't yield.

### 2. A browser that doesn't open is completely silent — `scripts/spellson.lic:150-159`

```ruby
respond "[spellson: #{url}]" unless Lich::WebUI.open_browser?
@browser = Lich::WebUI.open(page: @page, presentation: ..., geometry: ...)
@browser = nil unless @browser.respond_to?(:apply)
```

`BrowserLauncher.open` returns `false` when no Chrome/Edge can be found or spawned — on Linux it only looks at
three hardcoded Google Chrome paths (`browser_launcher.rb:18-22`), so a Firefox-only box fails here — and
`PresentedWindow.open` turns that into `nil` (`presented_window.rb:36-37`). Line 159 then quietly discards it.
The player sees no window, no URL and no error; the loop at line 317 spins at 2 Hz until they think to
`;kill spellson`.

lich-5's equivalent logs on exactly this value: `log(:warning, 'browser window failed to open; page is
available at the launch URL') if opened == false` (`script_scope/gtk/session.rb:881`).

The same three lines have a second problem: under `--webui-no-browser` (remote play, headless box) spellson
prints the URL *and then calls `Lich::WebUI.open` anyway*. `BrowserLauncher.open` does not consult
`Options.open_browser?` — its only refusal check is the spec-only `LICH_WEBUI_NO_BROWSER` env var
(`browser_launcher.rb:60-64`) — so a browser is spawned regardless. The reference implementation returns early:

```ruby
unless Lich::WebUI.open_browser?
  announce_launch_url(page, url)
  return true
end
```

Restructure to match: return after announcing the URL when `open_browser?` is false, and `respond` the URL when
`open` comes back falsy.

Confidence: confirmed.

### 3. What the browser reports and what it is given back are not the same rectangle — `spellson.lic:158`, `:260`, `:298`

`windowGeometry()` (`app.js:76-82`) reports **outer** size and **viewport** position:

```js
return JSON.stringify({ width: window.outerWidth, height: window.outerHeight,
                        position: [window.screenX, window.screenY] });
```

Both halves are stored and fed back as if they were something else.

*Size.* `fitWindow` treats the geometry facility's width/height as the **inner** size and adds the chrome back
(`app.js:1670-1684`): `inner = max(geometry.width, natural.width)`, then
`resizeTo(inner + (outerWidth - innerWidth), ...)`. `resizeTo` sets the outer size, so each launch ends one
chrome-width and one chrome-height larger than the last, and that larger value is what gets saved.
`page.fitted` only suppresses this within a single window's life; every `;spellson` opens a fresh one.

*Position.* `window.screenX`/`screenY` are the **viewport's** offset from the screen, while Chrome's
`--window-position` places the **window frame**. Feeding one back as the other shifts the window down by about
the title-bar height on every launch. `b7c4e61d` is what introduced this half — it is the right fix for
"position is never restored", but it inherits the mismatch.

Neither half is spellson's invention: `WebUILauncher` stores outer geometry too ("Persists validated
outer-window geometry", `webui_launcher/window_geometry_store.rb:10`) and passes the same hash to the same
opener (`webui_launcher.rb:82`), so the real fix probably belongs in lich-5 — either have `fitWindow` stop
adding chrome to a value that already includes it, or have `windowGeometry()` report a consistent rectangle.
Raising it here because this PR's description promises the size and position round-trip, and they drift.

One spellson-only wrinkle while you're in there: `load_settings` substitutes `[0, 0]` for an absent position
(line 188), and line 158 now passes that unconditionally, so a profile with nothing saved opens pinned to the
screen's top-left instead of wherever the OS would have put it. The launcher's store keeps `position: nil` for
"unset" (`window_geometry_store.rb:14`) and `geometry_arguments` skips the flag when it is nil
(`browser_launcher.rb:119`) — worth matching.

Confidence: suspected — traced through both sides of the contract, but not observed in a running browser. One
launch / move / relaunch cycle with the numbers written down settles both halves.

## Minor

### 4. Constants are redefined on every re-run — `scripts/spellson.lic:64-98`

`module SpellsOn` with thirteen constants plus `Row = Struct.new(...)` is defined at script top level, so a
second `;spellson` in the same Lich session re-initializes all of them and Ruby warns on each. This repo fixed
exactly this two commits before the PR branched, in `7fd0b971` (`fix(briefcombat): v1.0.5 prevent constant
redefinition on subsequent runs`, #2481), which switched `STATE_TAG_PAIRS`/`MONO_ON_PATTERN` to `||=` in a
module with the same shape. Same treatment applies here.

### 5. `@browser` is captured and then never used — `scripts/spellson.lic:155-159`, `:162-165`

The comment above `window_presentation` says "The GTK window was kept above everything, every tick", and the GTK
version did re-assert `@window.keep_above = true` on every pass of the loop (base `spellson.lic:144`). The new
code applies it once, when `PresentedWindow#adopt` fires, and then never again — `@browser` is assigned, checked
for `apply`, and dropped. `PresentedWindow#apply` exists for this ("a script re-applies after a setting changes
by calling `apply` and nothing else", `presented_window.rb:20-21`) and re-reads the callable each time.

Either call `@browser&.apply` in the run loop and make the comment true, or drop `@browser` and the
`respond_to?` dance and reword the comment to say it's applied once at open.

### 6. Fatal errors leak the page and the browser window — `scripts/spellson.lic:327-331`

```ruby
rescue StandardError => e
  respond ...
  SpellsOn.current_window = nil
  raise
```

`current_window` is cleared *before* re-raising, so `before_dying` (line 336) finds nil and skips both
`save_settings` and `destroy`. Nothing else calls `terminate_owner` — outside `lib/webui/` the only callers in
lich-5 are the GTK shim and the launcher, for their own owners. The page stays in the registry and the browser
window stays open showing a dead page for the rest of the Lich session. Move the nil-ing after cleanup, or just
let `before_dying` do the work and drop it from the rescue.

### 7. The changelog entry is written for a reviewer, not a player — `scripts/spellson.lic:18-25`

Eight lines mentioning "contract 2.17 bar and label layers", "composite", "Settings keys" and "no theme signal".
The repo's convention is explicit that entries are short, non-technical and describe the user-visible change
(`CLAUDE.md`, and see `eloot.lic` v2.11.6-v2.11.9 for the house voice). Most of this belongs in the PR body,
where it already is.

More importantly it omits the two things a player actually needs to know: this version needs a recent Lich, and
it needs Chrome or Edge installed. The death-decay bar changing colour is worth a line too.

## Nits

- `spellson.lic:47` — `require File.join(LIB_DIR, 'webui.rb')` raises an unrescued `LoadError` on a Lich with no
  `lib/webui.rb`, so the friendly contract message at lines 50-57 never fires on that path. In practice
  `;repo` already stops this: `check_lich_requirement` (`repository.lic:364-395`) reads the `required: Lich >=
  5.22.0` header off the downloaded file and `download_file` (`:1995-2000`) deletes it and refuses to install
  unless `--force`, and both install paths (`:2616`, `:2668`) go through that. So this only bites a `--force`
  install or a hand-copied file. If you want it closed anyway, `exit unless Script.require_lich_version!` above
  line 47 does it in one line and reuses the header you already declare — `combat_stats.lic:50-56` uses exactly
  that, with the same reasoning in its comment. Otherwise it may be worth a word noting lines 50-57 are for the
  developer case (webui present but older than 2.17), since that is the only case that reaches them.
- `spellson.lic:53` — the message says "too old" for any contract that isn't `2.x, x >= 17`, including a future
  3.0. Cosmetic, but "incompatible" would age better.
- `spellson.lic:76` — `SIDE_ALLOWANCE = 56` doesn't reconcile with the comment above it (16 of window border +
  16 of padding per side = 48). If the extra 8 is a fudge factor, say so.
- `spellson.lic:186`, `:217` — `Settings.load`/`Settings.save` are legacy no-ops in current Lich
  (`settings.rb:537-546`); writes persist on assignment. Harmless, carried from v1.0.0.
- `spellson.lic:210-218` — now that the write actually happens, a window drag can flush twice a second, and each
  flush is three `Settings[...]=` calls, each of which serialises and writes the whole settings blob
  (`settings.rb:379`). Cheap guard: skip `save_settings` when the three values match what was last written.

## Pre-existing (not introduced here)

`spellson.lic:242` — `@durations[spell.num] ||= spell.duration.to_f`. `Spell#duration` is an alias for
`time_per_formula` (`lib/common/spell.rb:930`), which returns the *formula source string* from
`effect-list.xml`, not a number. Across that file's 208 non-stackable self-cast durations, 30 begin with a
non-numeric character and `.to_f` to `0.0` (spell 106: `"(Spell[106].known? ? 120 : 20) + Spells.minorspiritual"`;
1003/1006/1007/… : `"Spellsong.timeleft"`), and 18 more `.to_f` to a wrong number (9655: `"10/60.0"` → 10.0,
not 0.167). So those bars sit permanently empty or permanently near-empty.

The PR is right that "the row maths is unchanged" — v1.0.0 had the same call and divided by it, producing
`Infinity` instead. Not this PR's bug, but the new guard (`duration.positive? ? ... : 0.0`, line 243) converts a
loud failure into a silent one, and the fix is small: `Spell#time_per` evaluates the formula and returns a Float.
Worth a follow-up either way.

## Open questions

- **The geometry timer is reset by every render.** `acceptRender` clears `page.geometryTimer` and
  `trackWindowGeometry` restarts a fresh 500 ms interval with `previous` re-sampled (`app.js:84-94`, `1886`).
  The run loop refreshes roughly every 500 ms + work, because `Spell#timeleft` decays continuously
  (`spell.rb:354-366`) so `rows == @rows` is false on essentially every pass. That makes the geometry timer fire
  in a sliver just before each render. **Settles it:** when you resized the window in game, did the new size
  actually come back on the next launch, or did you verify that path only through the page builder?
- **Cost of the 2 Hz full re-render.** Every 500 ms the whole tree is rebuilt, re-validated (3 layers per spell)
  and the browser does `page.element.replaceWith(next)` (`app.js:1907`). **Settles it:** with ~30 spells up, is
  CPU where you'd expect? If not, gating the refresh on `row.text` changing (roughly once a minute) rather than
  on `fraction` would cut it by ~100x at the cost of choppier bars.
- `spellson.lic:12` says `required: Lich >= 5.22.0`, but `LICH_VERSION` on the WebUI branch is still `5.21.0`.
  **Settles it:** is 5.22.0 the planned release number for the WebUI stack?

## What this PR gets right

- **The palette conversion is exact.** All sixteen CSS colour names from the old light palette map to the
  correct hex triples — `violet` → `EE82EE`, `burlywood` → `DEB887`, `lightsalmon` → `FFA07A`, and so on — and
  the set is byte-for-byte the same keys the GTK light branch had, including the absence of `'12'`.
- **The 6666 → 66 observation is correct.** `(spell.num).to_s[0..-3]` on `"6666"` is `"66"`, so the old
  `'6666'` key genuinely never matched and death decay always fell through to the red default. Good catch,
  correctly explained in the PR body rather than smuggled in.
- **The `b7c4e61d` dirty-flag handoff is the right shape**, not a lock-it-and-hope. Marking under the lock,
  reading-and-clearing under the lock, and clearing again inside the save means neither a lost flush nor a
  double write; `flush_settings` being documented as script-thread-only is what makes it reviewable.
- **Two v1.0.0 leaks are fixed in passing**: `@durations` is pruned against the active list (line 228) where the
  old `duration` hash grew forever, and division by a zero duration is guarded (line 243) where the old code
  produced `Infinity`.
- **The contract is used correctly.** Layer kinds as strings, `emphasis: 'strong'`, RGBA tone records, `value`
  clamped into `0.0..1.0` with a `finite?` guard first, 3 layers per spell against a 512-layer bound — I checked
  each against `contract.json` and the validator spec; the tree validates.
- **`terminate_owner`, not `service.stop`**, with the reasoning in a comment. Easy thing to get wrong in a
  shared-service world.
- The comment at lines 42-46 explaining the `LIB_DIR` case-sensitivity hazard is the kind of thing that saves
  the next person an afternoon.

## Coverage notes

- Full clone of `elanthia-online/scripts`; both PR commits fetched and the head file read in full, not just the
  diff. The `46442072` → `b7c4e61d` delta was reviewed line by line on its own.
- lich-5 was read from the `pr-1657` ref (contract 2.19), since the WebUI is not on `main` — the ref the PR
  body points at as #1648-#1655 plus fixes. Anything asserted about `Lich::WebUI`, `app.js`, `Settings`,
  `Script` or `Spell` is quoted from that tree. If the merged form of the stack differs, findings 1-3 should be
  re-checked against it.
- `bundle exec rubocop` with the repo config on the new head: clean, no offenses. `ruby -c`: Syntax OK.
  Ruby 4.0.5.
- No runtime verification: no Lich session, no game connection, no browser. Finding 3 and the first open
  question are the ones that most want a live check.
- `effect-list.xml` duration analysis was done over the copy in this repo at `scripts/effect-list.xml`, which is
  the file lich-5 downloads at `spell.rb:188`.

Nisugi:
Thanks, this was a thorough one. Everything is addressed at `f771dd17`; the two client-side halves are on lich-5's `webui/review-2-fixes`.

**Major 1, service start.** `start_service` now has the shape of the shim's: `running?` short circuit, direct start from the default group, otherwise a parked thread moved before it is released, `ThreadError` rescued. Applied to `;map` (#2479) in the same pass, since it had the same three lines.

**Major 2, browser open.** Under `--webui-no-browser` the URL is announced and nothing is spawned; when `open` comes back falsy the URL is announced with a note that no window could be opened.

**Major 3, geometry round trip.** Both halves were real and you had the cause right.
- *Size* is fixed on the lich-5 side: `windowGeometry()` now reports `innerWidth`/`innerHeight`, which is what `fitWindow` takes a declared geometry to be. Confirmed in game before the fix on `;map`: placed, reopened, grew once, reopened, held (clamped by `availHeight`, which is why three quick launches earlier showed nothing).
- *Position*: the client now calls `moveTo` on the same first fit that sizes the window, since Chrome ignores `--window-position` once running exactly as it ignores `--window-size`. `screenX`/`screenY` versus the frame is the one part I have not measured; one launch/move/relaunch with numbers written down is the next in-game check.
- The unset case is `nil` now, offered to neither the opener nor the facility, so a first run opens where the OS puts it.

**Minors.** Constants use `||=`; keep-above is re-applied every pass of the loop so the comment is true; the fatal-error rescue leaves `current_window` for `before_dying`; the changelog is three player-facing lines, including the Lich and browser requirements and the death-decay colour.

**Nits.** "incompatible"; the side allowance is now explained by what is actually inside the reported (inner) width, 48; `Settings.load`/`save` gone; an unchanged geometry report writes nothing. I left `require_lich_version!` out on purpose: `LICH_VERSION` on the WebUI branch is still 5.21.0 and it would refuse to run on the very Lich it targets. 5.22.0 is the planned number for the stack, and `;repo` already gates on the header.

**Pre-existing, taken anyway.** `@durations` is filled from `Spell#time_per`, so formula durations (106, bard songs, 9655) fill correctly. Good catch, and the guard no longer hides it.

**Open questions.** The geometry-timer reset was a real hole: the tracker re-sampled its baseline on every render, so a page redrawn faster than 500 ms swallowed its own resize. Fixed in the client: the baseline is the last geometry *reported* and survives renders, with a harness case (five renders 200 ms apart, one report). On cost: the text changes every second anyway (`as_time` shows seconds), so the render rate is inherent to a countdown; ~30 spells is 90 layers and validation is well under a millisecond, and the browser swap is one element. I have not measured CPU with 30 up, but nothing here is heavier than the shim's ProgressBar path was.

mrhoribu (commented):
# PR Review: feat(spellson.lic): v2.0.0 rewritten from GTK to Lich's WebUI (elanthia-online/scripts#2482)

**Reviewed:** 1 file, +353/-110 at `f771dd17` on 2026-09-17 (third pass; earlier passes at `46442072` and
`b7c4e61d`). Based on: full clone of `scripts`, plus `lich-5` read at `origin/webui/review-2-fixes` —
`lib/webui/**`, `lib/webui/assets/app.js`, `app.css`, `lib/common/webui_launcher*`,
`lib/common/script_scope/gtk/session.rb`, `lib/common/settings*`, `lib/common/script.rb`, `lib/common/spell.rb`,
`spec/webui_client/cases/window.test.mjs`. Rubocop with the repo config: clean. `ruby -c`: clean. No
runtime/browser verification.
**Verdict:** Approve with comments
**Summary:** `f771dd17` answers every Major and Minor from the previous pass, and I verified each against the
code rather than the commit message — including the two lich-5 commits it leans on. What's left is one
factually wrong line in the player-facing changelog, a robustness caution around the new `time_per` call, and a
cross-repo merge-ordering dependency worth stating out loud.

## Verified fixed in `f771dd17`

- **Service start no longer races** (`spellson.lic:174-197`). The starter is created parked on `gate.pop` and
  only released after `ThreadGroup::Default.add`, so the accept loop cannot inherit this script's group. The
  `service.server.running?` short-circuit and the `ThreadError` rescue are both there, and
  `Lich::WebUI.service` / `Service#server` / `Server#running?` are all real public API
  (`webui.rb:40`, `service.rb:15`, `server.rb:101`). Faithful to `start_service` in
  `script_scope/gtk/session.rb:276-297`.
- **`--webui-no-browser` is honoured and a failed open is reported** (`spellson.lic:148-163`). The early return
  means nothing is spawned; `@browser` is nil exactly when `PresentedWindow.open` returned nil
  (`presented_window.rb:36-37`), so the `unless @browser` message fires on precisely the no-Chrome/Edge path.
- **Geometry, both halves.** Position: nil until the browser reports one, so a first run is placed by the OS,
  and the facility and the opener each carry a position only when there is one (`:157-158`, `:306-311`). Size:
  fixed on the lich-5 side in `85eed237` — `windowGeometry()` now reports `innerWidth`/`innerHeight`, which is
  what `fitWindow` takes the facility to be, so the chrome is no longer added twice. `SIDE_ALLOWANCE ||= 48`
  is consistent with that: `#pages { padding: 16px }` (`app.css:24`) is 32 across, leaving 16 for a scrollbar.
  The client also moves the window now (`583a1d86`), which is what makes the saved position actually stick —
  Chrome ignores `--window-position` when it is already running.
- **Constants use `||=`** (`:62-100`), matching `7fd0b971`'s fix in `briefcombat.lic`.
- **Keep-above is re-applied every pass** via `present_window` (`:199-201`, `:371`), so the comment about the
  GTK window is true again. `PresentedWindow#apply` returns false cheaply where there is no hwnd, so this costs
  nothing off Windows.
- **The fatal-error path cleans up** — `current_window` is left set for `before_dying` (`:379`).
- **The changelog is player-facing** and now names the two things that matter operationally (Lich 5.22+,
  Chrome or Edge). See the one correction below.
- **Nits**: "incompatible" wording, the `SIDE_ALLOWANCE` comment, and suppressing a write when nothing changed
  (`:253-265` — `next` inside `synchronize` releases the lock correctly, and the shallow `@saved_settings.dup`
  is safe because `window_position` is always replaced with a new array, never mutated in place).
- **Pre-existing duration bug**: `spell.time_per` in place of `spell.duration.to_f` (`:286-292`) is the right
  accessor — `duration` is an alias for `time_per_formula`, the formula's source text (`spell.rb:930`), while
  `time_per` evaluates it (`spell.rb:338-347`).

## Minor

### 1. The changelog names a spell that was never affected — `scripts/spellson.lic:23-25`

```
    - Bars for a few spells whose duration is a formula (Spirit Warding,
      bard songs) used to sit empty; they fill properly now.
```

Spirit Warding I (101) is `span='stackable'` in `effect-list.xml`, so `spell.stacks` is true and `row_for` took
the `STACK_CEILING` branch — it never read `duration` and its bar was never empty. Same for Spirit Barrier (102)
and Spirit Defense (103), if those were the intended family.

Bard songs are a correct example: Fortitude Song (1003), Song of Luck (1006) and the rest carry no `span`, so
they are non-stackable, and their duration formula is `Spellsong.timeleft` — which `.to_f` read as 0.

Pick a replacement from the spells that actually hit it. Spell 106 (Disease Resistance's neighbour,
`"(Spell[106].known? ? 120 : 20) + Spells.minorspiritual"`), 916, and the Sunfist/Voln timers in the 99xx range
are all real cases. Or just say "some spells" and skip the naming — this is the one line in the file a player
will read, and it shouldn't assert something checkable and wrong.

## Nits

- `spellson.lic:290` — `spell.time_per` runs `proc { eval(formula) }.call` (`spell.rb:343`) on data from
  `effect-list.xml`. That widens the exposure: lich itself only calls `time_per` on the cast path
  (`spell.rb:575-577`), i.e. for spells *you* cast, whereas spellson now evaluates it for every active spell —
  including ones cast on you by others and the 9xxx game timers, whose formulas may never otherwise run. It is
  called inside `update`, inside the main loop, with `run`'s rescue above it re-raising, so one formula that
  raises takes the whole window down. `@durations[spell.num] ||= (spell.time_per.to_f rescue 0.0)` costs a
  line and degrades to the old empty-bar behaviour instead of killing the script.
- `spellson.lic:226-236` — one correction to my own earlier nit: `Settings.save` is genuinely a no-op
  (`settings.rb:537-540`, returns `:noop`), but `Settings.load` is **not** — it aliases `refresh_data`
  (`:542-545`), which deletes the cache key and re-reads from the database (`:386-392`). Dropping it is still
  fine here, since the cache is keyed per script and scope and only this script writes that key, so what's
  cached is what's stored. Flagging it because I lumped the two together last pass and that was imprecise.
- `spellson.lic:199` / `:203` — `present_window` and `window_presentation` sit two lines apart and read almost
  identically. `reapply_presentation` or similar would save the next reader a double-take.

## Open questions

- **Is the reported position the same origin the client moves to?** `windowGeometry()` reports
  `[window.screenX, window.screenY]`, which MDN defines as the **viewport's** offset from the screen, while
  `fitWindow`'s new `window.moveTo(x, y)` (`app.js`, from `583a1d86`) moves the **window**. If those differ by
  the title bar on Chrome, a reopened window walks down the screen by that much each time — the same shape as
  the size ratchet `85eed237` just removed, one layer down. The client harness stubs `moveTo` and asserts the
  call (`spec/webui_client/cases/window.test.mjs:72-87`) but never defines `screenX`/`screenY`, so it doesn't
  exercise the round trip. **Settles it:** open the window, drag it somewhere, close and reopen twice — does it
  come back where you left it, or a title bar lower each time? This is a lich-5 question, but spellson is where
  it shows.
- **Merge ordering.** `SIDE_ALLOWANCE = 48` is only correct with `85eed237`, and the saved position only sticks
  with `583a1d86`. Both are on `origin/webui/review-2-fixes`, not on lich-5 `main`. If 5.22.0 ships without
  them, this script gets the growing window back *and* under-allows its width by the border. **Settles it:**
  are both commits inside the 5.22.0 cut?
- **Cost of the 2 Hz full re-render.** Unchanged from the last pass: every 500 ms the tree is rebuilt,
  re-validated (3 layers per spell) and the browser does `page.element.replaceWith(next)`. **Settles it:** with
  ~30 spells up, is CPU where you'd expect?

## What this PR gets right

- **The fixes were made at the right layer.** The window-growth bug could have been papered over in spellson
  with a fudge constant; it was fixed in `windowGeometry()` where the wrong rectangle was being reported, which
  fixes the launcher and `;map` at the same time. Same for the position: `--window-position` was the wrong
  mechanism, and `moveTo` on first fit is the right one.
- **`start_service` was copied in full, not approximated** — the parked-thread gate, the `running?`
  short-circuit and the `ThreadError` rescue are all present. The easy version of this fix is to move the
  `ThreadGroup` call earlier and hope; this one closes the race by construction.
- **The palette conversion is exact.** All sixteen CSS colour names map to the correct hex triples, and the key
  set matches the GTK light branch including the absence of `'12'`.
- **The 6666 → 66 observation is correct.** `(spell.num).to_s[0..-3]` on `"6666"` is `"66"`, so the old
  `'6666'` key never matched and death decay always fell through to the red default.
- **The settings write path is honest about its threading.** `settings_changed` / `flush_settings` /
  `save_settings` split, with `flush_settings` documented script-thread-only and the dirty flag read-and-cleared
  under the lock, is a structure a reviewer can check rather than trust.
- **The contract is used correctly** — layer kinds as strings, `emphasis: 'strong'`, RGBA tone records, `value`
  clamped into `0.0..1.0` behind a `finite?` guard, 3 layers per spell against a 512-layer bound.

## Coverage notes

- Full clone of `elanthia-online/scripts`; all three PR commits fetched, the `b7c4e61d` → `f771dd17` delta read
  line by line, and the head file read in full.
- lich-5 read at `origin/webui/review-2-fixes`, which carries `85eed237` and `583a1d86`. Everything asserted
  about `Lich::WebUI`, `app.js`, `app.css`, `Settings`, `Script` or `Spell` is quoted from that ref. Neither
  commit is on lich-5 `main` yet — see the merge-ordering question.
- `bundle exec rubocop` with the repo config on the new head: clean. `ruby -c`: Syntax OK. Ruby 4.0.5.
- No runtime verification: no Lich session, no game connection, no browser. The position open question is the
  one that most wants a live check.
- The `span='stackable'` claim about Spirit Warding was checked against `scripts/effect-list.xml` in this repo,
  which is the file lich-5 downloads at `spell.rb:188`.
- **Withdrawn from earlier passes:** the `LoadError` finding. `repository.lic` gates on the `required:` header
  (`check_lich_requirement`, `:364-395`; enforced in `download_file`, `:1995-2000`, on both install paths at
  `:2616` and `:2668`), so a Lich below 5.22.0 never receives the file. I verified the gate fires for this
  script's exact header.

Nisugi:
Thanks for the third pass, and for checking the lich-5 commits rather than taking the message's word. All three comments are in at `83ed4439`.

**Changelog.** You are right, Spirit Warding stacks and never read the formula. The line now names bard songs and "some society and game timers", which are the real cases (1003/1006 via `Spellsong.timeleft`, 916, the 99xx timers).

**`time_per` exposure.** Guarded: a formula that raises leaves that bar empty (`0.0`, the old behaviour) instead of unwinding through `update` and ending the script. The comment says why spellson is the only place these formulas run for spells you did not cast.

**`Settings.load`.** Noted, and thanks for the correction; dropping it stays right for the reason you give.

**Naming.** `present_window` is `reapply_presentation`.

**Open questions.**
- *Position origin.* Still the one thing I have not measured; the in-game drag / close / reopen twice is queued and I will post the numbers here. If `screenX`/`screenY` turns out to be the viewport, the fix is in `windowGeometry()` next to the size fix, not here.
- *Merge ordering.* Both commits (`583a1d86`, `85eed237`) are on `webui/review-2-fixes`, the single fixes branch on top of the #1648-#1655 stack, and are meant to land with it; a third client commit (the geometry tracker keeping its baseline across renders, the "timer reset" hole you found last pass) sits on the same branch locally. I will make sure the ordering is stated on the lich-5 side.
- *Cost.* Unchanged from my last answer; not measured with 30 up yet.

